PR #7, opened 2014-08-09
Support SNI.

ruricolist:
Support for SNI, adapted from https://github.com/akamai/cl-http2-protocol/blob/master/ssl.lisp.

### DIFF
--- a/ffi.lisp
+++ b/ffi.lisp
@@ -208,6 +208,66 @@
   (larg :long)
   (parg :pointer))
 
+(cffi:defcfun ("SSL_CTX_callback_ctrl" ssl-ctx-callback-ctrl)
+    :long
+  (ctx ssl-ctx)
+  (cmd :int)
+  (fp :pointer))
+
+(defconstant +SSL_TLSEXT_ERR_OK+ 0)
+(defconstant +SSL_CTRL_SET_TLSEXT_SERVERNAME_CB+ 53)
+(defconstant +SSL_CTRL_SET_TLSEXT_SERVERNAME_ARG+ 54)
+(defconstant +SSL_CTRL_SET_TLSEXT_HOSTNAME+ 55)
+(defconstant +TLSEXT_NAMETYPE_host_name+ 0)
+
+(defmacro ssl-ctx-set-tlsext-servername-callback (ctx cb)
+  `(ssl-ctx-callback-ctrl ,ctx +SSL_CTRL_SET_TLSEXT_SERVERNAME_CB+ ,cb))
+
+(defmacro ssl-ctx-set-tlsext-servername-arg (ctx arg)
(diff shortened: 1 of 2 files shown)


richsalz:
Just a brief note (it's late, I'm tired:)  There are no plans to change the OpenSSL SNI-related API at this point.  It is, sadly, not documented.  I'm a member of the openssl dev team; see https://github.com/openssl/openssl/commits/master  What are the specific questions or concerns, I'll try to answer them.

avodonosov:
@richsalz thanks.

Lets limit the question to client-side:

Is it enough for client to just `SSL_set_tlsext_host_name` or client should also `SSL_CTX_set_tlsext_servername_callback` ? 

I ask because we see in the curl sources (https://github.com/bagder/curl/blob/785395b07e01eb9085655d6ef311e34c2fca2224/lib/vtls/openssl.c) callback is not specified, but in the OpenSSL s_client.c (https://github.com/openssl/openssl/blob/b548a1f11c06ccdfa4f52a539912d22d77ee309e/apps/s_client.c) the callback is used. It seems like s_client.c only uses this callback to print a message to debug output -  whether or not the server acknowledges SNI extension - and nothing else.

richsalz:
Yes, you have it exactly right.  the set_tlsext sets it in the clientHello, and the callback is used to see if the server responds with the value in its response.

avodonosov:
@richsalz Thanks.

So, in scope of this ticket we will only add SSL_set_tlsext_host_name when initializing the request.

richsalz:
Glad to help.  Does this mean I helped get something in to common lisp?  Way cool!

avodonosov:
Yes, it does :)

avodonosov:
So, I've added the `hostname` parameter to `cl+ssl:make-ssl-client-stream`. Tested it with the following code

```common-lisp

(pushnew "/home/anton/prj/cl+ssl/cl-plus-ssl/" asdf:*central-registry*)
(ql:quickload :cl+ssl)
(ql:quickload :usocket)

(let* ((socket (usocket:socket-connect
                "sni.velox.ch"
                443
                :element-type '(unsigned-byte 8)))
       (ssl-stream (cl+ssl:make-ssl-client-stream (usocket:socket-stream socket)
                                                  :hostname "sni.velox.ch"))
       (char-stream (flexi-streams:make-flexi-stream ssl-stream
                                                     :external-format '(:utf-8 :eol-style :crlf)))
       (reply-buf (make-string 1000)))
  (unwind-protect
       (progn
         (format char-stream "GET / HTTP/1.1~%")
         (format char-stream "Host: sni.velox.ch~%~%")
         (finish-output char-stream)
         (let ((reply)))
         (read-sequence reply-buf char-stream)
         reply-buf)
    (usocket:socket-close socket)))

```

Depending on the `hostname` parameter it returns different results.

@deadtrickster lets discuss other your suggestions - validation and context creation - in separate issues (you've created #8 for validation)

Thanks all for participation in this ticket.

richsalz:
This is cool; great to see more SNI support.  Thanks.

PR #11, opened 2015-04-28
Add cipher selection to client streams

tacosundae:
Adding support for client selection of ciphers. 

I needed this to access a service behind an F5 BigIP load-balancer which completely chokes on client hellos with too many cipher specs. The default "ALL" cipher list is way over that. 

Nothing special, is was all there already, just needed to be used.


avodonosov:
Thanks

PR #15, opened 2015-11-22
replace tabs with spaces

deadtrickster:
Lets start with this.
As you can see in the diff tabs and spaces were mixed even in the same line!


avodonosov:
thanks, merged

PR #17, opened 2015-11-22
ssl-ctx-set-cipher-list

deadtrickster:
Usage example:

```lisp
(ssl-ctx-set-cipher-list ssl-ctx
                         (format nil
                                 "ECDHE-RSA-AES256-GCM-SHA384:~
                                  ECDHE-RSA-AES256-SHA384:~
                                  ECDHE-RSA-AES256-SHA:~
                                  ECDHE-RSA-AES128-GCM-SHA256:~
                                  ECDHE-RSA-AES128-SHA256:~
                                  ECDHE-RSA-AES128-SHA:~
                                  ECDHE-RSA-RC4-SHA:~
                                  DHE-RSA-AES256-GCM-SHA384:~
                                  DHE-RSA-AES256-SHA256:~
                                  DHE-RSA-AES256-SHA:~
                                  DHE-RSA-AES128-GCM-SHA256:~
                                  DHE-RSA-AES128-SHA256:~
                                  DHE-RSA-AES128-SHA:~
                                  AES256-GCM-SHA384:~
                                  AES256-SHA256:~
                                  AES256-SHA:~
                                  AES128-GCM-SHA256:~
                                  AES128-SHA256:~
                                  AES128-SHA"))
```

### DIFF
--- a/ffi.lisp
+++ b/ffi.lisp
@@ -196,6 +196,14 @@
   (ssl ssl-pointer)
   (str :string)
   (type :int))
+(cffi:defcfun ("SSL_CTX_set_cipher_list" ssl-ctx-set-cipher-list%)
+    :int
+  (ctx :pointer)
+  (ciphers :pointer))
+(defun ssl-ctx-set-cipher-list (ctx ciphers)
+  (cffi:with-foreign-string (ciphers* ciphers)
+    (when (= 0 (ssl-ctx-set-cipher-list% ctx ciphers*))
+      (error 'ssl-error-initialize :reason "Can't set SSL cipher list" :queue (read-ssl-error-queue)))))
 (cffi:defcfun ("SSL_CTX_use_certificate_chain_file" ssl-ctx-use-certificate-chain-file)
     :int
   (ctx ssl-ctx)


avodonosov:
Honestly, I was in doubts about this commit: all other functions in ffi.lisp just plain ffi definitions, but new ssl-ctx-set-cipher-list checks return code raises a condition. Maybe it's a better approach but things become not "coherent". Another doubt: why ssl-error-initialize, not just ssl-error ?

But eventually I decided to not block your work with these minor issues.

deadtrickster:
Hey, thank you for the feedback
I totally understand your doubts however I think `ssl-ctx-set-cipher-list` is a little bit different from others since int future it wiill be a part of ffi public interface. Moreover even now if user wants to override *global-context* and wants to set cipher list (but hey, if he overrides context he already knows cl+ssl insane defaults like "ALL" so he definitely will do this) he have to call `ssl-ctx-set-cipher-list` directly. Hence error checking. 

Other candidates for public API
`ssl-ctx-new`  `ssl-ctx-set-verify-depth` `ssl-ctx-set-verify` etc.

Sometimes I think about more highlevel context-related constructs. Maybe even wrapping SSL context in its class and using it with macro `with-ssl-context`.

Why `ssl-error-initialize`?  `ssl-ctx-set-cipher-list` is a context initialization operation.
Also  `ssl-set-cipher-list` fails converted to `ssl-error-initialize` in `make-ssl-client-stream` ;-)

avodonosov:
Hi, thanks for the new pull requests.

> Why ssl-error-initialize? ssl-ctx-set-cipher-list is a context initialization operation.
> Also ssl-set-cipher-list fails converted to ssl-error-initialize in make-ssl-client-stream ;-)

Yes, I saw ssl-set-cipher-list failures converted to ssl-error-initialize, but it's at concrete call site, where this concrete caller thinks of ssl-set-cipher-list call as of initialization. If we put the return status handling into ssl-ctx-set-cipher-list itself, we don't know in what situations it will be called - maybe as initialization, or maybe changing cipher list in the middle of work. That was my thought. OK, not so important, I don't mind too much.

About public API. I think ffi.lisp code (+ some minor things like reload) can be given a separate package - e.g. opensssl-ffi - and all the functions will be public. Just a package of plain FFI bindings to OpenSSL API. There is nothing to hide there, all the bindings we collected can be used by anyone.

And more lispy / high-level thinks, like stream creation uses that openssl-ffi package.

PR #20, opened 2015-11-23
Fix fd leak in close when ssl-stream-socket is raw fd

deadtrickster:
For example drakma calls make-ssl-client-stream
with raw fd
https://github.com/edicl/drakma/blob/master/util.lisp#L332

not sure about error type though


avodonosov:
Type ERROR is OK I think, esp. taking into account there are lot of other (ERROR ...) invokations

PR #8, opened 2014-09-25
Check host/email/ip

deadtrickster:
Hostname/email verification code ported from OpenSSL 1.0.2 beta2.
Also includes verify callback enabling more detailed validation errors reporting. 

PS I want to write tests for matching code but it looks like I have to choose testing library by myself. I'm considering fiveam or maybe lift. What you think?

PR #21, opened 2015-11-24
missing close-fd 

deadtrickster:
As you probably noticed I slowly break https://github.com/cl-plus-ssl/cl-plus-ssl/pull/8 into pieces 
Here https://github.com/cl-plus-ssl/cl-plus-ssl/pull/20 I forgot to add close-fd import. Sorry about that.

### DIFF
--- a/ffi.lisp
+++ b/ffi.lisp
@@ -50,6 +50,11 @@
 
 ;;; Function definitions
 ;;;
+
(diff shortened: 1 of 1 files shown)


avodonosov:
Yes, I noticed. Thanks a lot for that.

PR #24, opened 2015-12-05
Make-context

deadtrickster:
Provides `with-global-context` macro and `make-context` function

Can help with problems like described here: 
https://github.com/edicl/drakma/issues/50

Examples:

```lisp
(cl+ssl:with-context 
    ((cl+ssl:make-context :verify-location "/etc/ssl/certs"))
  (drakma:http-request "https://google.com" :force-ssl t))
```

```lisp
(cl+ssl:with-context 
    ((cl+ssl:make-context :verify-mode cl+ssl::+ssl-verify-none+))
  (drakma:http-request "https://incomplete-chain.badssl.com/"))
```

```lisp
(cl+ssl:with-context 
    ((cl+ssl:make-context :verify-mode cl+ssl::+ssl-verify-peer+))
  (drakma:http-request "https://incomplete-chain.badssl.com/"))

SSL verify error: 20 X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY
```

```lisp
(cl+ssl:with-context 
    ((cl+ssl:make-context :verify-mode cl+ssl::+ssl-verify-peer+ :verify-callback nil))
  (drakma:http-request "https://incomplete-chain.badssl.com/"))

A failure in the SSL library occurred on handle #.(SB-SYS:INT-SAP #X7FFE6825EC40) (return code: 1).
SSL error queue:
error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed
```


avodonosov:
After a very superficial look: I think better name instead of `make-global-context` would be `make-context` or `new-context` or even `context`

deadtrickster:
Actually it was `make-context`. I changed it just before creating pull request. Why?
Because `*ssl-global-context*` -> `with-global-context` -> `make-global-context`.

I don't think cl+ssl:(make-)context is a good name for this particular function since there is always two contexts (SSL_CTX_new and SSL_ctx). And this naming can be confusing.

avodonosov:
No no, I think you probably forgetting (due to late hour?) your original, correct, reasoning. 

SSL_CTX_new and SSL_ctx are not two context. SSL_CTX_new creates new context, it can create one, two or as many contexts as we want. SSL_ctx is a slot of ssl session storing a context earlier created by SSL_CTX_new.

 `make-context` is just a convenience wrapper around SSL_CTX_new, that's all.

deadtrickster:
yeah you are right, any other thoughts? I'll change the name right now

deadtrickster:
done!

package.lisp
```diff
@@ -23,4 +23,11 @@
            #:ssl-set-global-default-verify-paths
            #:ssl-error-verify
            #:ssl-error-stream
-           #:ssl-error-code))
+           #:ssl-error-code
+
+           #:+ssl-verify-none+
+           #:+ssl-verify-peer+
+           #:+ssl-verify-fail-if-no-peer-cert+
+           #:+ssl-verify--client-once+
```
avodonosov: two dashes - is it typo?

context.lisp
```diff
@@ -0,0 +1,103 @@
...
+  ((location :initarg :location))
+  (:documentation "Unable to find verify locations")
+  (:report (lambda (condition stream)
+              (format stream "Unable to find verify location. Path: ~A" (slot-value condition 'location)))))
+
+(defun validate-verify-location (location)
+  (handler-case
+      (cond
+        ((uiop:file-exists-p location)
+         (values location t))
```
avodonosov: We need to add explicit dependency on :uiop in the .asd

avodonosov: We will also need to document new public API in the doc (index.html)

avodonosov: We will also need to document new public API in the doc (index.html)

avodonosov:
We will also need to document new public API in the doc (index.html)

avodonosov:
Makes sense to add `ssl-ctx-free` to the public API - the users who call `make-context` will want to free the context.

avodonosov:
@deadtrickster do you think it would be convenient to have a `auto-free` option in the `with-global-context` macro?

context.lisp
deadtrickster: it's a part of asdf. afaik it loaded implicitly

package.lisp
deadtrickster: yes

deadtrickster:
@avodonosov 
maybe not ssl-ctx-free but 
```lisp
(declaim (inline free-context))
(defun free-context (context)
  (ssl-ctx-free context))
```
?

`auto-free` - why not? but only if it set to nil by default

documentation - news section is extremely outdated, as well as copyright years and holders btw :-)
I will add doc entry after `ensure-initialized`.

context.lisp
avodonosov: It's a dependency of ASDF, but not in all ASDF versions. UIOP exists as a separate module and is available in Quicklisp.

avodonosov:
> maybe not ssl-ctx-free but
> 
> (declaim (inline free-context))
> (defun free-context (context)
>   (ssl-ctx-free context))
> ?

It's too much IMO.

I don't wan't to hide real OpenSSL API, and introduce our different API. I think it's better to think of cl+ssl as a plain ffi bindings to OpenSSL + some helpers.

> auto-free - why not? but only if it set to nil by default

nil by default, or course, I fully agree.

I wasn't sure about auto-free because it will complicate the invocation format of `with-global-context`; if the cases when auto-free is used are rare - probably most users will reuse their context more than once - then the majority of cases will suffer from complicated `with-global-context` while not having benefits of auto-free.

> documentation - news section is extremely outdated, as well as copyright years and holders btw :-)
> I will add doc entry after ensure-initialized.

Thanks.

Yes, there are outdated parts, in particular I am considering to remove the news section to avoid maintenance. But not now.

deadtrickster:
> I think it's better to think of cl+ssl as a plain ffi bindings to OpenSSL + some helpers.

Well given that cl+ssl implements gray streams I don't think it is plain ffi + some helpers
I'm sure coherent naming (esp without performance penalty) is way more important. If you call `make-context` why you should call `ssl-ctx-free` but not `free-context`? 

> auto-free

I played with it and you know what? It can kill you if you'll try to reuse freed context (basically on sbcl at least you are done and have to restart sbcl)! And as you rightly assume contexts indeed intended to be long-lived
Btw this is closely related to previous citation: should cl+ssl really punish its users that hard if they are trying to reuse freed contexts?

I remember when we met last time on #lisp we agreed that cl+ssl actually has two parts - plain ffi bindings and a lot of helpers ;-). Or I misunderstood something? Lets make it clear now since next step I plan is actually introduce wrapper around ssl-ctx structure with which I can at least test if it was freed, initialized or not (as I said earlier it just kills entire lisp instance in case of an error). If you think it is too much then I'll continue to contribute ffi related-stuff (there so much to do too) but all this highlevel things will go to my separate repo.

deadtrickster:
Changed some defaults, added documentation

avodonosov:
As discussion becomes more general, first about this particular pull reqest:

1. No need for reader conditional in ASDF dependencies `#-asdf3 :uiop` (we just depend on :uiop, no matter how it is provided - either loaded already by ASDF3 or any other way. Also reader conditionals in asdf system definitions create problems for systems like Quicklisp where you need to know what are the dependencies or particular system).

2. Please let's just use `ssl-ctx-free`. Yes, we call 'make-context` to create context, but we know `make-context` is just a helper around `ssl-ctx-new` and returns normal SSL context. More on this below.

auto-free - we can add it later if we feel it's convenient.

I meant auto-free for the cases as in your examples:
```common-lisp
(cl+ssl:with-global-context (cl+ssl:make-context :verify-location "/etc/ssl/certs")
  (drakma:http-request "https://google.com" :force-ssl t))
```
To save user from freeing this temporary, one-time context explicitly. More correct example:

```common-lisp
(cl+ssl:with-global-context (cl+ssl:make-context :verify-location "/etc/ssl/certs")
  (prog1
    (drakma:http-request "https://google.com" :force-ssl t)
   (ssl-ctx-free cl+ssl:*global-context*))
```
 BTW, if someone reuses freed context, the problems are understandable - `SSL_ctx_free` is most likely calls C function `free` on the context pointer, so we can expect usual problems if this pointer is used afterwards.

And I wouldn't say cl+ssl punishes users. This isn't a error prone API - it's unlikely anyone will reuse context if he consciously called `free` on the context. I think more often / probable error will be when people forget to call `free`.

>  I remember when we met last time on #lisp we agreed that cl+ssl actually has two parts - plain ffi bindings and a lot of helpers ;-). Or I misunderstood something? 

Yes, yes.

> Well given that cl+ssl implements gray streams I don't think it is plain ffi + some helpers

Streams, yes, but I still regard it as an utility. BTW, implementing custom stream takes that lot of code partially because of incomplete factorization of stream API: we don't distinguish between the API stream provides to user (gray streams API) and the API to implement when one wants to create custom stream. Compare to Java - streams, readers provide quite rich interfaces, but to supply you own source of data it's enough to override just one method - `read`.

> Lets make it clear now since next step I plan is actually introduce wrapper around ssl-ctx structure with which I can at least test if it was freed, initialized or not (as I said earlier it just kills entire lisp instance in case of an error). If you think it is too much then I'll continue to contribute ffi related-stuff (there is so much to do too) but all these highlevel things will go to my separate repo.

What I mean when say cl+ssl is ffi bindings + utilities, is that I want to avoid formulating and implementing another fully fledged SSL API, which hides OpenSSL. It will be hard to make it convenient and fully expressive, and I afraid there will be not enough free time and attention to do it well. And I doubt the final API will be significantly better than OpenSSL API. 

But most importantly, it will confuse users: now user deals with complex OpenSSL API; then it will deal with two complex API: OpenSSL + our lispy API. which talk about the same things but in different terms.

If you will create your own CTX structure, you will need to duplicate all the functions which use contexts (initialization, setting parameters, inheriting from context and other usages). One thing will lead to another and we will end-up with a whole pile of new functions.

It's necessary to see really significant benefits and have lot of resources before entering this route.

When I talk about utilities, I mean it's better to hve our lispy functionality transparent, fully compatible and composable with basic OpenSSL API.

But all that is lyrics. Actually, if consider next steps I think the most valuable gift for the community would be a patch to drakma which implements the behaviour specified in drakma API in case when cl+ssl is used. And this work will give very valuable information about how to shape out the cl+ssl API.

But let's first finish with this pull request.

avodonosov:
Let's make `ssl-ctx-free` public

deadtrickster:
About #-asdf - I just googled how others do it. 
I don't want to fight with you over `free-context` mostly because reevaluating my own use cases I returned `:auto-free`.

deadtrickster:
added (hopefully concise) documentation entry for `with-global-context`

avodonosov:
Thank you very much for new features and for being so responsive during the review.

 It's a great contribution - enables many convenient use cases, while preserving backward compatibility.

avodonosov:
BTW, I forgot one important thing: do we use any new OpenSSL functions which present only in new versions of OpenSSL? (As I learned from your pull request #8, some features like verifying host/email/ip were absent in old OpenSSL and people needed to implement it themselves; later these features were added). Do we now depend on some similar, newly introduces OpenSSL API?

deadtrickster:
good question. recently a lot of stuff was deprecated and unfortunately openssl usually shipped very outdated. For example `tls_method` introduced that effectively replaces `ssl_v23_method` and no_sslv2|no_sslv3` flags i.e. it allows TLS only. rsa_callback deprecated too.

Now I don't think hostname verification  should rely on openssl implementation. At least until we implement reliable openssl-version dependent switches. But even with these switches there is a great possibility of subtle difference between old-lisp  and new-openssl implementations.

I have battle-tested erlang implementation, I'll just backport it and cover with tests.

avodonosov:
I mean in this pull request, new functions we use, like `SSL_CTX_set_verify`, or the ciphers used in `+default-cipher-list+` - are they backwards compatible, or they are only work with new OpenSSL? (I think unlikely, but just to check).

> I have battle-tested erlang implementation [of hostname verfication], I'll just backport it and cover with tests.

Doesn't your pull request #8 provide this?

deadtrickster:
`SSL_CTX_set_verify` is backwards compatible. +default-cipher-list+ should be too unless openssl EXTREMELY outdated.

Yep my pull request provide this however it was tested and used by me only. While https://github.com/deadtrickster/ssl_verify_hostname.erl is widely used and therefore I can be sure in its algorithm and tests.

avodonosov:
Listen, I can't reproduce the verify-peer failure from examples in the description. Instead of the error I get normal HTTP reply:

```
(cl+ssl:with-global-context 
    ((cl+ssl:make-context :verify-mode cl+ssl::+ssl-verify-peer+))
  (drakma:http-request "https://incomplete-chain.badssl.com/"))

=>
"<!DOCTYPE html>
<html>
<head>
  <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">
  <link rel=\"shortcut icon\" href=\"/icons/favicon-orange.ico\"/>
  <link rel=\"apple-touch-icon\" href=\"/icons/icon-orange.png\"/>
  <title>incomplete-chain.badssl.com</title>
  <link rel=\"stylesheet\" href=\"/style.css\">
  <style>body { background: rgb(243, 121, 46); }</style>
</head>
<body>
<div id=\"content\">
  <h1 style=\"font-size: 8vw;\">
    incomplete-chain.<br>badssl.com
  </h1>
</div>

</body>
</html>
"
200
((:SERVER . "nginx/1.6.2 (Ubuntu)") (:DATE . "Sun, 06 Dec 2015 23:43:37 GMT") (:CONTENT-TYPE . "text/html") (:CONTENT-LENGTH . "506") (:LAST-MODIFIED . "Fri, 04 Dec 2015 02:46:28 GMT") (:CONNECTION . "close") (:ETAG . "\"5660fe84-1fa\"") (:CACHE-CONTROL . "no-store") (:ACCEPT-RANGES . "bytes"))
#<URI https://incomplete-chain.badssl.com/>
#<FLEXI-STREAMS:FLEXI-IO-STREAM #x302002FCD07D>
T
"OK"
```

avodonosov:
I am trying on quicklisp 2015-09-24, CCL 1.9 on linux.

deadtrickster:
Hmm at first I was thinking about my drakma patches, but:

``` lisp 

(cl+ssl:with-global-context 
 ((cl+ssl:make-context :verify-mode cl+ssl::+ssl-verify-peer+))
 (usocket:with-client-socket (socket stream "incomplete-chain.badssl.com" 443
                                     :element-type '(unsigned-byte 8))
                             (let* ((ssl-stream (cl+ssl:make-ssl-client-stream stream
                                                                               :hostname  "incomplete-chain.badssl.com"))
                                    (char-stream (flexi-streams:make-flexi-stream ssl-stream
                                                                                  :external-format '(:utf-8 :eol-style :crlf)))
                                    (reply-buf (make-string 1000)))
                               (unwind-protect
                                   (progn
                                     (format char-stream "GET / HTTP/1.1~%")
                                     (format char-stream "Host: incomplete-chain.badssl.com~%~%")
                                     (finish-output char-stream)
                                     (read-sequence reply-buf char-stream)
                                     reply-buf)
                                 (close ssl-stream)))))
```

also signals
```
SSL verify error: 20 X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY
```

avodonosov:
This usocket example behaves the same for me

deadtrickster:
ok let me install ccl too

avodonosov:
By "the same" I mean signals failre - "the same" as your result

avodonosov:
Only the drakma example differs - doesn't fail

deadtrickster:
oh thank you :-) so looks like drakma issue. lets dive deeper into this when hostname verification merged

PR #30, opened 2015-12-12
Verify hostname

deadtrickster:
This should replace https://github.com/cl-plus-ssl/cl-plus-ssl/pull/8

TODO:
- [ ] Complete ASN1 String types handling
- [ ] Null-pointer checks
- [ ] Error types


avodonosov:
Haven't you forget to add "test/certs/" directory?

deadtrickster:
@avodonosov nope, it's here: https://github.com/deadtrickster/cl-plus-ssl/tree/6ce0516bccec4c461896fec794bd3277bcaa8079/test/certs

deadtrickster:
ASN1 string validation references:
 - https://github.com/digitalbazaar/forge/blob/909e312878838f46ba6d70e90264650b05eb8bde/js/asn1.js
 - http://www.obj-sys.com/asn1tutorial/node128.html
 - https://github.com/deadtrickster/ssl_verify_hostname.erl/blob/master/src/ssl_verify_hostname.erl
 - https://golang.org/src/encoding/asn1/asn1.go?m=text

deadtrickster:
looks like I need help with ecl build. On one machine it passed. On another it fails but differently (for fresh setup like on Travis and when restarting). Moreover this simple check:

``` lisp 
(cl+ssl::validate-and-parse-wildcard-identifier "*.qwe.com" "ww.qwe.com")
T
""
".qwe.com"
T
```
that obviously passes on sbcl, ccl, clisp fails on ECL with cl+ssl just loaded (returns NIL). However if I copy-paste `check-single-wildcard` AND `validate-and-parse-wildcard-identifier` definitions to REPL after cl+ssl loaded this test passes.

avodonosov:
Maybe ti's an ECL bug. I don't know now. If so, we can disable ECL.

This pull request, is it not ready for merge, should I wait? I see the TODO items in description are all finished.

avodonosov:
BTW, if we say that ffi bindings can become a separate module, maybe it's better to keep all ffi bindings in ffi.lisp, instead of keeping part of them in x509.lisp?

And yes, I now see the test/certs - somehow didn't notice them initially.

avodonosov:
@dkochmanski, do you think https://github.com/cl-plus-ssl/cl-plus-ssl/pull/30#issuecomment-164293951 is an ECL bug?

We are testing relatively simple, self containing code and its behavior on ECL differs from other lisps and isn't always repeated on ECL.

dkochmanski:
I'll check this with https://github.com/deadtrickster/cl-plus-ssl/tree/verify_hostname and my develop build.

dkochmanski:
[package cl+ssl].
    ;;; Internal error:
    ;;;   ** Reader error in file #<input stream #P"/home/jack/quicklisp/local-projects/cl-plus-ssl/ffi.lisp">, position 5410:
    ;;; 0 is an extra argument for the #X readmacro.

dkochmanski:
In ffi.lisp:122 we have
    (defconstant +SSL-VERIFY-FAIL-IF-NO-PEER-CERT+ #0x02)
is it valid syntax? I'll remove the leading zero for testing purposes

dkochmanski:
hrm, what branch should I check? grepping the verify_hostname branch directory for "validate" doesn't bring anything

    grep -rni "validate"

dkochmanski:
yes, it's ECL bug. I'll work on it tomorrow. Seems, that find behaves strangely in function check-single-wildcard from verify-hostname.lisp . Thanks for noticing me. I'll report back here when I'll fix it.

deadtrickster:
@dkochmanski Hey thanks for being so responsive!
@avodonosov I added NULL checks, probably this can be merged now

dkochmanski:
Hey! I've fixed this in develop branch. It was a problem with the compiler macro. If you are interested, you may check it here: https://gitlab.com/embeddable-common-lisp/ecl/issues/199 .

avodonosov:
@dkochmanski , cool, thanks.

deadtrickster:
@avodonosov any chance this will be merged soon?

src/x509.lisp
```diff
@@ -0,0 +1,360 @@
...
+
+(defgeneric decode-asn1-string (asn1-string type))
+
+(defun copy-bytes-to-lisp-vector (src-ptr vector count)
+  (cffi:with-pointer-to-vector-data (dst-ptr vector)
+    (memcpy
+     dst-ptr
+     src-ptr
+     count)))
+
```
avodonosov: strange formatting, why not `(memcpy dst-ptr src-ptr count)`

src/x509.lisp
```diff
@@ -0,0 +1,360 @@
...
+(cffi:defcfun ("strlen" strlen)
+    :int
+  (string :string))
+
+(cffi:defcfun (memcpy "memcpy") :pointer
+  (dest :pointer)
+  (src  :pointer)
+  (count :int))
+
```
avodonosov: I'm afraid memcpy and strlen won't be available on all platforms - standard C library is not installed everywhere.

avodonosov:
> ASN1 string validation references:
> 
> https://github.com/digitalbazaar/forge/blob/909e312878838f46ba6d70e90264650b05eb8bde/js/asn1.js
> http://www.obj-sys.com/asn1tutorial/node128.html
> https://github.com/deadtrickster/ssl_verify_hostname.erl/blob/master/src/ssl_verify_hostname.erl
> https://golang.org/src/encoding/asn1/asn1.go?m=text

Makes sense to add such comment to the source code

src/x509.lisp
deadtrickster: Luckily both strlen and memcpy seems to be already used by OpenSSL itself.

deadtrickster:
Hey, thanks for the review, I corrected formatting, added references comment and commented your strlen/memcpy note.

src/x509.lisp
avodonosov: But it doesn't mean strlen and memcpy are necessary exported for linking to other programs.

deadtrickster: I'm not sure I understand you here. Could you give real-world example or propose something?

avodonosov: It wont work when somebody builds OpenSSL linking to standard C library statically. In this case the binary executable image of OpenSSL library doesn't have strlen and memcpy as symbols available for linking to other programs.

A solution - to use lisp functions to count bytes and copy them.

deadtrickster: Yeah I always forget about windows :-(

I'll rewrite this using cffi:mem-ref

deadtrickster: I'm completely replacing strlen with (cffi::foreign-string-length data :encoding :ascii)

deadtrickster:
So strlen replaced with cffi::foreign-string-length and memcpy replaced with loop (dotimes)

src/x509.lisp
```diff
@@ -0,0 +1,358 @@
...
+    (if (asn1-printable-string-p bytes)
+        (babel:octets-to-string bytes :encoding :ascii)
+        (error 'invalid-asn1-string :type '+v-asn1-printablestring+))))
+
+(defmethod decode-asn1-string (asn1-string (type (eql #.+v-asn1-utf8string+)))
+  (let* ((data (asn1-string-data asn1-string))
+         (length (asn1-string-length asn1-string))
+         (strlen (cffi::foreign-string-length data :encoding :ascii)))
+    (if (= strlen length)
+        (cffi:foreign-string-to-lisp data)
```
avodonosov: it's better to be explicit about encoding - ` (cffi:foreign-string-to-lisp data  :encoding :utf-8)`

avodonosov: Also, taking into account that "In general it cannot be assumed that the data returned by ASN1_STRING_data() is null terminated " [1], IMHO we need to pass :count : `(cffi:foreign-string-to-lisp data :encoding :utf-8 :count length)`

[1] - https://www.openssl.org/docs/manmaster/crypto/ASN1_STRING_dup.html

src/x509.lisp
```diff
@@ -0,0 +1,358 @@
...
+ASN1 string validation references:
+ - https://github.com/digitalbazaar/forge/blob/909e312878838f46ba6d70e90264650b05eb8bde/js/asn1.js
+ - http://www.obj-sys.com/asn1tutorial/node128.html
+ - https://github.com/deadtrickster/ssl_verify_hostname.erl/blob/master/src/ssl_verify_hostname.erl
+ - https://golang.org/src/encoding/asn1/asn1.go?m=text
+|#
+(defgeneric decode-asn1-string (asn1-string type))
+
+(defun copy-bytes-to-lisp-vector (src-ptr vector count)
+  (declare (type (simple-array (unsigned-byte 8)))
```
avodonosov: Do you mean `(type (simple-array (unsigned-byte 8)) vector)` ?

src/x509.lisp
```diff
@@ -0,0 +1,358 @@
...
+  (< byte #x80))
+
+(defun asn1-iastring-p (bytes)
+  (declare (type (simple-array (unsigned-byte 8)) bytes)
+           (optimize (speed 3)
+                     (debug 0)
+                     (safety 0)))
+  (every #'asn1-iastring-char-p bytes))
+
+(defmethod decode-asn1-string (asn1-string (type (eql #.+v-asn1-iastring+)))
```
avodonosov: Hm, why `#.` ?

deadtrickster: I remember there were some problems on lisps other than sbcl, probably clisp. that is why respective defconstants wrapped in eval-when, but there is a possibility that #. was used before eval-when introduced and therefore obsolete now

avodonosov: I've studied this, and have a theory what problem you could have: according to CLHS `defonstant` is not required to evaluate the value-form at compile time, it only required to inform compiler about the constant name; and `defmethod` evaluates the eql-specializer-form at compiler time (when expanding the `defmethod` macro).

I think this might be the problem you experienced. But I've tested on CLISP, CLISP compiles this without a problem

```common-lisp
(defconstant c 1)
(defgeneric g (param))
(defmethod g ((param (eql c)))
  "hello")
```
So maybe it was another CL.

Anyway, it that was the problem, `eval-when` should fix that.

So we can avoid the `#.` - it's not pretty to have the in the code.

Moreover, I am going to ask you to move the ffi definitions to the ffi.lisp. And this will even make `eval-when` unnecessary, because if x509.lisp depends on ffi.lisp, ASDF will compile and _load_ ffi.lisp before compiling x509.lisp; thus we will have the defconstant fully executed.

avodonosov:
@deadtrickster , first of all, sorry for the slow review - please don't take it as if I'm not interested to have it merged, I just don't have enough free time after work. Also this pull request is relatively complex - a lot to learn (ASN.1, ASN.1 string types)

I'd like to ask you to move the ffi definitions from x509.lisp to ffi.lisp. We currently have this separation, and the file name ffi.lisp tells the code reader this file contains ffi definitions. I understand one can question usefulness of such separation; maybe it's more convenient for you to edit x509 related foreign definitoins "in place", right near your lispy x509 wrappers. But if do so, we should refactor everything and remove ffi.lisp at all.

ffi.lisp name gives reader a signal "ffi is here". Thinking it's not useful is one thing. But leaving the ffi.lisp while using different principle of code organization, ffi.lisp name becomes a _wrong_ signal, it still tells "ffi is here" while it's false, ffi is not only there, but in other places too. And when we mix two classifications: some code is separated as ffi/not ffi, and other code is divided as x509/not x509, it's like for example say that mobile phones maybe classified as samsung phones, or black phones.

It's one argument. Another argument is that I hope to move the plain OpenSSL FFI bindings into a separate package; and have lispy wrappers in their own packages.

Also, moving the defs to ffi.lisp will it unnecessary to use `#.` with ffi contstans in x509.lisp.

avodonosov:
@deadtrickster Another question: why in x509.lisp you want not only convert ASN.1 strings, but also to validate their encoding? Anyway, this validation is not complete.

src/x509.lisp
```diff
@@ -0,0 +1,355 @@
...
+  (declare (type (simple-array (unsigned-byte 8)) bytes)
+           (optimize (speed 3)
+                     (debug 0)
+                     (safety 0)))
+  (every #'asn1-teletex-char-p bytes))
+
+(defmethod decode-asn1-string (asn1-string (type (eql #.+v-asn1-teletexstring+)))
+  (let* ((data (asn1-string-data asn1-string))
+         (length (asn1-string-length asn1-string))
+         (strlen (cffi::foreign-string-length data :encoding :ascii)))
```
avodonosov: `data` is not guaranteed to be null terminated

src/x509.lisp
```diff
@@ -0,0 +1,355 @@
...
+    (if (= strlen length)
+        (let ((bytes (asn1-string-bytes-vector asn1-string)))
+          (if (asn1-teletex-string-p bytes)
+              (cffi:foreign-string-to-lisp data :encoding :ascii)
```
avodonosov: `:count` is needed

src/x509.lisp
```diff
@@ -0,0 +1,355 @@
...
+  (declare (type (simple-array (unsigned-byte 8)) bytes)
+           (optimize (speed 3)
+                     (debug 0)
+                     (safety 0)))
+  (every #'asn1-iastring-char-p bytes))
+
+(defmethod decode-asn1-string (asn1-string (type (eql #.+v-asn1-iastring+)))
+  (let* ((data (asn1-string-data asn1-string))
+         (length (asn1-string-length asn1-string))
+         (strlen (cffi::foreign-string-length data :encoding :ascii)))
```
avodonosov: `data` is not guaranteed to be null terminated

src/x509.lisp
```diff
@@ -0,0 +1,355 @@
...
+    (if (= strlen length)
+        (let ((bytes (asn1-string-bytes-vector asn1-string)))
+          (if (asn1-iastring-p bytes)
+              (cffi:foreign-string-to-lisp data :encoding :ascii)
```
avodonosov: `:count` is needed

deadtrickster:
Ok, so null checks were removed and ffi stuff moved to ffi.lisp.

deadtrickster:
ping

avodonosov:
@deadtrickster what about encoding validation? Why do we need this?

If validating encoding is important, OpenSSL should do it, if not important, why would we do it?
(I took some investigation into OpenSSL sources, I don't see OpenSSL validating encoding).

I see in the go x509 parse you refer they validate encodings, but we don't parse the certificate, OpenSSL does it for us.

On the one hand we are know of for performance, up to even declaiming character testing functions inline (which wont work when we pass function as an argument to `every` and so on). On the other hand, why would we spend time on checking character ranges at all?

src/x509.lisp
```diff
@@ -0,0 +1,225 @@
...
+  (declare (type (simple-array (unsigned-byte 8)) bytes)
+           (optimize (speed 3)
+                     (debug 0)
+                     (safety 0)))
+  (every #'asn1-iastring-char-p bytes))
+
+(defmethod decode-asn1-string (asn1-string (type (eql +v-asn1-iastring+)))
+  (let ((bytes (asn1-string-bytes-vector asn1-string)))
+    (if (asn1-iastring-p bytes)
+        (babel:octets-to-string bytes :encoding :ascii)
```
avodonosov: This would require to declare :babel as an explicit dependency in our ASDF system. But as we already depend on flexi-streams, we could use `flexi-streams:octets-to-string`

deadtrickster:
OK, I added babel as explicit dependency, also removed unnecessary inlines (I hope I just forgot to remove in first place).
Regarding validation - the more and more I'm exploring this topic the more libs that _do_ validation I find. Honestly I just don't care what openssl does (unless it already does some sort of validation) since all this ASN stuff is standardized and here I'm just following the spec.

avodonosov:
Thanks, I will take a look, but some time later - the following next days I'm busy.

01.03.2016, 11:53, "Ilya Khaprov" <notifications@github.com>:

deadtrickster:
Hey, if you so busy maybe you just give me write access? I don't think my code all that bad. Especially now when we integrated with Travis.

deadtrickster:
ping

avodonosov:
We can decode ASN1 strings using `ASN1_STRING_to_UTF8` :

```common-lisp

(cffi:defcfun ("SSLeay" ssleay)
    :long)

(cffi:defcfun ("ASN1_STRING_to_UTF8" asn1-string-to-utf8)
    :int
  (out** :pointer)
  (asn1-string :pointer))

;; Up to OpenSSL 1.0.2 CRYPTO_free had only one argument
(cffi:defcfun ("CRYPTO_free" crypto-free-old)
    :void
  (addr :pointer))

;; After 1.0.2 crypto_free has 3 arguments, and is included into public API
(cffi:defcfun ("CRYPTO_free" crypto-free-new)
    :void
  (addr :pointer)
  (file :pointer)
  (line :int))

;; This value is never dealocated (we pass it as the FILE parameter
;; to CRYPTO_free, which expects a C string of static storage class)
(defparameter +unknown-file+
  (cffi:foreign-string-alloc "cl+ssl: source file is unknown"))

(defun crypto_free (pointer)
  ;; if version > 1.0.2
  (if (> (logand (ssleay) #xfffff000)
         #x10002000)
      (crypto-free-new pointer +unknown-file+ 0)
      (crypto-free-old pointer)))

(defun decode-asn1-string (asn1-string)
  (cffi:with-foreign-objects ((buf :pointer) (buf* :pointer))
    (setf (cffi:mem-ref buf* :pointer) buf)
    (let* ((length (asn1-string-to-utf8 (cffi:mem-ref buf* :pointer)
                                        asn1-string)))
      (if (< length 0)
          (error 'ssl-error-call :message "Error decoding ASN1 string"
                 :queue (read-ssl-error-queue))
          (unwind-protect
               (cffi:foreign-string-to-lisp (cffi:mem-ref buf :pointer)
                                            :count length :encoding :utf-8)
            (crypto-free (cffi:mem-ref buf :pointer)))))))

```

avodonosov:
@deadtrickster of course your code is not that bad; and you generously devote energy and knowledge - without you there would be no progress in this area. But code review reveals important nuances (remember at least strlen/memcpy).

We will commit all this soon.

deadtrickster:
ping?

eugeneia:
Just want to chime in and say thanks to the great work you guys are doing! Both @deadtrickster for the relentless effort, and @avodonosov for being responsibly throrough. :heart:

avodonosov:
@eugeneia well, thanks, I'm of course kind of blocking the work because want the best solution but don't have time. The next thing to do is to email OpenSSL and ask them not remove the SSLeay function (which the've renamed to OpenSSL_version which is not only a breaking change, but also doesn't allow you to conditionally adopt to other braking changes they may with reasonable effort. I's already around a half of a year that I'm planning to do that - not very responsive I must admit.

libre-man:
What is needed to get this merged? As this issue (together with revoked certs) is the reason that ssl isn't used in quicklisp (quicklisp/quicklisp-projects#1315). I would love to help if I could.

avodonosov:
@libre-man demonstrating that the suggested API is adequate by a drakma patch would help.

libre-man:
So checking this in drakma was quite easy (simply set the `ssl-check-verify-p` to `t`). However wouldn't it be better to refactor the code around verification somewhat? As the user of the library now has two different ways to enable verification:
1) When creating the context.
2) Setting the `cl+ssl:ssl-check-verify-p` function to non `NIL`.

However both options seem to do something else which is quite annoying, why not merge the two?

Furthermore the `hostname-verification-error` condition does not inherit from any existing condition. Wouldn't it be better if it would extend `ssl-error-verify`?

avodonosov:
@libre-man , thanks. 

Key question to me is how drakma API will look after integrating with the cl+ssl hostname verification feature. Could you please explain more detailed?

libre-man:
Nothing will change. At the moment drakma does not set `ssl-check-verify-p` to something so this function will not even activate. If `ssl-check-verify-p` would be set (as it should be) and the hostname of the found cert is wrong an error will be signaled.

Why would the API of drakma change? A certificate for another hostname is simply not valid so when drakma verifies SSL (`:verify` set to `:required`) it should also verify the hostname.

@avodonosov How would you feel about refactoring the code a bit so that `ssl-check-verify-p` would become obsolete and just setting the `verify-mode` when creating a context to something else than `SSL-VERIFY-NONE` would enable `ssl-check-verify-p`? 

One more thing: on line 96 of verify-hostname.lisp there is TODO still in the code. Would code like [this](https://gist.github.com/libre-man/12a4ce4b7218cca8ce80db475f0927a1) solve that todo @deadtrickster?

eugeneia:
To me it seems as if this should be a new silent default. I.e. after this lands all libraries using cl+ssl including Drakma should automatically verify certificates. An option to disable verification could be useful for certain testing scenarios, but I doubt it is desired by default.

libre-man:
@eugeneia The way it is implemented this is the case when the user of cl+ssl has set `ssl-check-verify-p` to not `NIL` and not `:UNSPECIFIED`.

deadtrickster:
Errors base type can be fixed of course.
 I don't see any reason why this should depend on drakma. Given no one cared for years.

eugeneia:
@libre-man Yes, and I am saying `ssl-check-verify-p` should (must?) be true by default. It should take extra effort to disable verification, not the other way around.

libre-man:
I agree with @eugeneia about the default value of `ssl-check-verify-p`. At the moment I think everything should work, however in the future it might be nice to offload these checks to OpenSSL if a supporting version is found.

A extra test to make sure the check is also done within a stream might be nice, so doing a webrequest like is done in sni.lisp.

avodonosov:
If drakma API doesn't change, we need to support drakma parametes `:verify nil`, `:verify :optional`, `:verify :required`. `max-depth`, `ca-file`, `ca-directory`. `certificate`, `key`, `certificate-password`.

libre-man:
At the moment it is not possible to supply a different hostname for verification. I don't see the usage of such a option, however we could simply do this by adding a key `verification-hostname` which we default to `hostname`. If this option is `nil` we could disable hostname verification.

Furthermore I don't think we should be thinking about what drakma might want. First of all there are other projects that use cl+ssl and I think we should focus on having sane and secure defaults.

I really don't know what `:verify :optional` should mean. In Allegro (where the option seems to come from) it means that it shouldn't throw an error when the server did not supply a certificate, however it should throw an error when the certificate supplied didn't verify. At the moment drakma doesn't do anything with this option for non allegro platforms. it could be nice to implement this, but I think it should be a different PR.

avodonosov:
`:verify :optional` is what 'ssl-check-verify-p' does now.

```common-lisp
(let ((cl+ssl::*ssl-check-verify-p* t))
  (drakma:http-request "https://expired.badssl.com/"))

=> SSL verify error: 10 X509_V_ERR_CERT_HAS_EXPIRED
```

Try the same for "https://null.badssl.com/" and it works without error.

Matching drakma requirements is a simple test for how sensible our API is, does it match at least one user. What other projects do you mean?

What you say about `verification-hostname` is convincing, we can add such parameter in the future if we want this level of control.

avodonosov:
Merged finally. I'm going to change asn1 decoding as in https://github.com/cl-plus-ssl/cl-plus-ssl/pull/30#issuecomment-196114925, but for now let it be in the current form.

@deadtrickster thanks for the solution, sorry for long merge.

avodonosov:
@deadtrickster, I'm experimenting with enabling (hostname) verification by default. The `verify-hostname` function turns out to have a strange protocol: returns verification result as true / false, but also signals an error for many verification failures.

deadtrickster:
AFAIR the idea is that it returns if verification can actually be done. Intended purpose is to answer question "is this hostname ok?". It signals errors if validation isn't possible i,e, something wrong with the certificate, invalid input basically.

avodonosov:
I've implemented the :verify parameter suitable for drakma (:required, :optional, nil) - eb97eb0.
Also enabled verification by default - 1ccf9d5 (however strong defender of backwards compatibility I am, this is a rare case when backwards compatibility might be violated this time - I think verification is what most of existing users expect)

PR #31, opened 2015-12-13
optimize locking-callback

deadtrickster:
Now each time it called (and this means REALLY frequently) cffi kindly converts foreign string to lisp for us. Given `file` not used it looks a bit redundant.


deadtrickster:
btw I believe recursive lock already fixed on sbcl. maybe it's time to to move on?

avodonosov:
about recursive locks - I don't see any benefits, the callback is called to lock / unlock, no need for recursive locks

PR #35, opened 2016-01-30
Use closesocket instead of close on Windows.

resttime:
Windows does not have the POSIX function close(int) to close a file descriptor, so trying to close an SSL stream results in an undefined alien error. Therefore, we replace the function with closesocket(int) to close the socket instead when using the Windows operating system.

```common-lisp
;; Closing the SSL stream breaks with an undefined alien error
;; because the close(int) function does not exist on Windows.

(ql:quickload "drakma")

(defun test (&optional (test-url "https://google.com"))
  (multiple-value-bind (input status) (drakma:http-request test-url
                               :want-stream T)
    (princ status)
    (close input)))
```


avodonosov:
Thanks!

PR #36, opened 2016-03-09
Add ssleay32.dll for windows

mattn:
msys2 provide ssleay32.dll instead of libssl32.dll

### DIFF
--- a/src/reload.lisp
+++ b/src/reload.lisp
@@ -25,7 +25,7 @@
   (cffi:use-foreign-library libcrypto))
 
 (cffi:define-foreign-library libssl
-  (:windows "libssl32.dll")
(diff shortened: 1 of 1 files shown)


avodonosov:
Hm, aren't they both needed: eay and ssl? See `define-foreign-library libeay32` below.

mattn:
> Hm, aren't they both needed: eay and ssl? See define-foreign-library libeay32 below.

Ah, right. I don't have libssl32.dll on msys2. But working openssl, or ruby/python's ssl stuff.

avodonosov:
I was confused, you commit is right (because ssleay = libssl, and libeay = libcrypto)

mattn:
Thank you

avodonosov:
Thank you too :+1:

PR #40, opened 2017-03-07
Use :long-long type at SSL_CTX_ctrl for 32bit environments.

fukamachi:
## Problem

On 32-bit environment, `CL+SSL:MAKE-CONTEXT` invokes a debugger:

```
? (cl+ssl:make-context)
> Error: The value 2181041151 is not of the expected type (SIGNED-BYTE 32).
> While executing: CL+SSL::SSL-CTX-CTRL, in process listener(1).
> Type :POP to abort, :R for a list of available restarts.
> Type :? for other options.
```

The type checking by CFFI failed because 2181041151 is not CFFI's :long type on 32-bit environments.

## Solution

The definition of `ssl-ctx-ctrl` which is called by `ssl-ctx-set-options` is here.

```
(cffi:defcfun ("SSL_CTX_ctrl" ssl-ctx-ctrl)
    :long
  (ctx ssl-ctx)
  (cmd :int)
  (larg :long)  ;; <= This type.
  (parg :pointer))
```

`(cffi:foreign-type-size :long)` returns 4 on 32-bit environments and perhaps it's small for `larg`.

Well, now, I'm a little bit confusing with that the [OpenSSL's document](https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_ctrl.html) says the `larg` type is `long`.
However, changing the type to `:long-long` fixes this problem.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -291,7 +291,7 @@ session-resume requests) would normally be copied into the local cache before pr
     :long
   (ctx ssl-ctx)
   (cmd :int)
-  (larg :long)
+  (larg :long-long)
   (parg :pointer))
 
 (cffi:defcfun ("SSL_ctrl" ssl-ctrl)


fukamachi:
Well, since there's no response, I felt I need to explain more.

I developed an HTTP client [Dexador](https://github.com/fukamachi/dexador), and I added SSL certificates verification for user's safety.

However, Dexador cannot verify SSL certificates at all because of this issue on 32-bit environment, like SBCL on Windows or 32-bit Clozure CL.
CL+SSL is the standard (probably only) library to use OpenSSL from Common Lisp, which means, **Common Lisp applications cannot verify SSL certificates on 32-bit environments**.

For solving this issue, I would help if there's something that I can.

FYI, Drakma, another de facto HTTP client, doesn't have this issue, because it never verify SSL certificates.

avodonosov:
Hi, sorry for the delay.

I don't understand the situation completely. What is 2181041151?

Not only OpenSSL docs, but also the source code declares larg as long: https://github.com/openssl/openssl/blob/75e314f2d573d4f984ff6a371be7a4966bf5f4c5/ssl/ssl_lib.c#L2098

In C the long type doesn't have precise lengths, it's only specified to be at least 32 bits. How does CFFI decides that long is a (SIGNED-BYTE 32)?

avodonosov:
I understood, 2181041151 is the result of (logior cl+ssl::+SSL-OP-ALL+ cl+ssl::+SSL-OP-NO-SSLv2+ cl+ssl::+SSL-OP-NO-SSLv3+) here: https://github.com/cl-plus-ssl/cl-plus-ssl/blob/a093bbb7bce3253d27b5350e2c7d12b9cdb1eb67/src/context.lisp#L102

Also note, that this is a 32 bit value: `=> 2181041151 (32 bits, #x82000BFF, #o20200005777, #b10000010000000000000101111111111)`

Only that it doesn't have place for a sign bit. Maybe unsigned long is better than long long here?

```common-lisp
(typep 2181041151 '(signed-byte 32))
=> NIL

 (typep 2181041151 '(unsigned-byte 32))
=> T
```

fukamachi:
Thanks for the reply.

> Maybe unsigned long is better than long long here?

That sounds better. I confirmed that `:unsigned-long` also solves the problem.

avodonosov:
Done. Thanks for the report.

fukamachi:
Thanks a lot.

PR #42, opened 2017-05-18
ssl-ctx-ctrl needs a ssl-context as first parameter.

mgi:
I don't know/understand why it used to work but it does not work anymore on
OpenBSD/LibreSSL.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -287,6 +287,10 @@ session-resume requests) would normally be copied into the local cache before pr
     ssl-pointer
   (file :string))
 
+(cffi:defcfun ("SSL_get_SSL_CTX" ssl-get-ssl-ctx)
+    ssl-ctx
+  (ssl ssl-pointer))
+
 (cffi:defcfun ("SSL_CTX_ctrl" ssl-ctx-ctrl)
     :long
   (ctx ssl-ctx)
--- a/src/streams.lisp
+++ b/src/streams.lisp
@@ -55,7 +55,7 @@
   (print-unreadable-object (object stream :type t)
     (format stream "for ~A" (ssl-stream-socket object))))
 
-(defclass ssl-server-stream (ssl-stream) 
+(defclass ssl-server-stream (ssl-stream)
   ((certificate
     :initarg :certificate
     :accessor ssl-stream-certificate)
@@ -174,14 +174,14 @@
 
 #+(and clozure-common-lisp (not windows))
 (defun install-nonblock-flag (fd)
-  (ccl::fd-set-flags fd (logior (ccl::fd-get-flags fd) 
+  (ccl::fd-set-flags fd (logior (ccl::fd-get-flags fd)
                      #.(read-from-string "#$O_NONBLOCK"))))
                      ;; read-from-string is necessary because
                      ;; CLISP and perhaps other Lisps are confused
-                     ;; by #$, signaling"undefined dispatch character $", 
-                     ;; even though the defun in conditionalized by 
+                     ;; by #$, signaling"undefined dispatch character $",
+                     ;; even though the defun in conditionalized by
                      ;; #+clozure-common-lisp
-                    
+
 #+(and sbcl (not win32))
 (defun install-nonblock-flag (fd)
   (sb-posix:fcntl fd
@@ -213,7 +213,7 @@
      (ssl-set-fd handle socket))
     (stream
      (ssl-set-bio handle (bio-new-lisp) (bio-new-lisp))))
-  (ssl-ctx-ctrl handle
+  (ssl-ctx-ctrl (ssl-get-ssl-ctx handle)
     +SSL_CTRL_MODE+
     +SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER+
     (cffi:null-pointer))


avodonosov:
> I don't know/understand why it used to work 

We can't commit if we don't know/understand. 

What exactly doesn't work? What is your use case?

mgi:
My use case is using drakma to access a https url.
On OpenBSD (-current) and sbcl 1.3.10, I have an unhandled memory fault in SSL_CTX_ctrl. But it seems to be because the wrong structure handle is pass to SSL_CTX_ctrl. The man page (https://www.openssl.org/docs/man1.1.0/ssl/SSL_CTX_ctrl.html) states that it should be a SSL_CTX structure handle and not a SSL structure handle (as it is the case in install-handle-and-bio called from make-ssl-client-stream).

avodonosov:
Do you use custom lisp library versions or Quicklisp? What quicklisp version?

The fix seems reasonable and corresponds to the OpenSSL docs, but the code should be failing long ago.

mgi:
Anton Vodonosov <notifications@github.com> writes:

> Do you use custom lisp library versions or Quicklisp? What quicklisp
> version?

I've seen the bug in the quicklisp version from 20161208 and then clone
master on github (which shows the same bug) to correct it.

> The fix seems reasonable and corresponds to the OpenSSL docs, but the
> code should be failing long ago.

Yes, this why I said earlier: "I don't understand why it used to
work". I have quickly checked and the two others call to ssl-ctx-ctrl
seems ok because they're called from wrapper that received a ssl-ctx
handle.
-- 
Manuel Giraud

avodonosov:
I have tried `(drakma:http-request "https://google.com/")` with and without your fix on SBCL 1.3.17 on Ubuntu x64, quicklisp 2017-02-27. Both versions work. 

I will accept your fix.

How I explain that the old version doesn't fail. The implementation of SSL_CTX_ctrl performs SSL_CTRL_MODE modification by bitwise-or updating memory at certain offset of the passed pointer: https://github.com/openssl/openssl/blob/42bd7a16d020de40ad166336fe60409aa9256f9b/ssl/ssl_lib.c#L2138

In our case SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER is 2, so 
```common-lisp
(ssl-ctx-ctrl (ssl-get-ssl-ctx handle)
    +SSL_CTRL_MODE+
    +SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER+
    (cffi:null-pointer)
```
means `ctx->mode |= 2` where `ctx` SSL handle was passed as the `ctx` parameters.
Depending on how openssl library is built and linked into the process it may or may not fire access violation. On my system it doesn't fail on your system it fails (I'm curious is because yours is OpenBSD).

BTW, I don't understand why SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER mode is set at all. After brief reading of the doc, to me it only looks useful in combination with partial writes mode SSL_MODE_ENABLE_PARTIAL_WRITE. But our code doesn't set SSL_MODE_ENABLE_PARTIAL_WRITE. Strange. (The doc: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_set_mode.html)

mgi:
Anton Vodonosov <notifications@github.com> writes:

> I have tried `(drakma:http-request "https://google.com/")` with and
> without your fix on SBCL 1.3.17 on Ubuntu x64, quicklisp
> 2017-02-27. Both versions work.
>
> I will accept your fix.

Thanks :-)

> means `ctx->mode |= 2` where `ctx` SSL handle was passed as the `ctx` parameters.
> Depending on how openssl library is built and linked into the process
> it may or may not fire access violation. On my system it doesn't fail
> on your system it fails (I'm curious is because yours is OpenBSD).

The OpenBSD folks have forked openssl since heartbleed. It is now named
"libressl" in their tree and modified independently. This bug doesn't
show in the previous release.

> BTW, I don't understand why SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER mode
> is set at all. After brief reading of the doc, to me it only looks
> useful in combination with partial writes mode
> SSL_MODE_ENABLE_PARTIAL_WRITE. But our code doesn't set
> SSL_MODE_ENABLE_PARTIAL_WRITE. Strange. (The doc:
> https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_set_mode.html)

Don't know but I didn't understand the same thing as you from the doc:
it seems that the adress of the write buffer is allowed to change.
-- 
Manuel Giraud

avodonosov:
Thinking more about this, I'm leaning towards just commenting out this fragment:

```common-lisp
(ssl-ctx-ctrl handle
    +SSL_CTRL_MODE+
    +SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER+
    (cffi:null-pointer))
```

It has never worked as intended, so we don't break anything. Implementing real context modification may be harmful - the context is a global object shared by all SSL instances. If we need to modify context then ` install-handle-and-bio` is not the place to do so. Probably the original author ( @lichtblau ) mean ssl-crl instead of ssl-ctx-ctrl.

mgi:
Anton Vodonosov <notifications@github.com> writes:

> Thinking more about this, I'm leaning towards just commenting out this fragment:
>
> ```common-lisp
> (ssl-ctx-ctrl handle
>     +SSL_CTRL_MODE+
>     +SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER+
>     (cffi:null-pointer))
> ```

FWIW, I've just tried this and it works for my simple use case too.
-- 
Manuel Giraud

avodonosov:
Done. @mgi, thanks for tracking down and reporting the problem.

kat-co:
FWIW [this](https://github.com/cl-plus-ssl/cl-plus-ssl/pull/42#issuecomment-304822501) worked for me too. Ubuntu 16.04 + Guix + SBCL 1.3.7 + Libressl 2.5.4 and `#<SYSTEM cl+ssl / cl+ssl-20170403-git / quicklisp 2017-05-16>` from Quicklisp. I suppose my issue is that the quicklisp version doesn't have this fix yet, so I had to manually comment out these lines.

PR #52, opened 2017-07-03
Fix SSL_OP_NO_SSLv2 constant value.

### LINKED ISSUE #50: SSL connection errors on Windows when using cl+ssl:make-context
Connecting to a server with `cl+ssl:make-ssl-client-stream` by itself works correctly, but creating a context and wrapping the call in `cl+ssl:with-global-context` causes the connection to error out (at least on CCL -- on SBCL, it hangs waiting for the connection to become readable). This happens on Windows 7, although there's one [report](https://github.com/fukamachi/dexador/issues/33#issuecomment-300043472) that it _doesn't_ happen on Windows 10 (I'll try to confirm this after work).

This causes problems for drakma tip, and the current quicklisp version of dexador (probably others too).

Test case:
````
(ql:quickload '(:cl+ssl :usocket))

(defun connect (host port)
  (let ((sock (usocket:socket-connect host port))
        (ctx (cl+ssl:make-context :verify-mode cl+ssl:+ssl-verify-none+
                                  :verify-location :default)))
    (cl+ssl:with-global-context (ctx :auto-free-p t)
      (cl+ssl:make-ssl-client-stream (usocket:socket-stream sock)
                                    :hostname host))))

(defun connect-stream (host port)
  (let ((sock (usocket:socket-connect host port)))
    (cl+ssl:make-ssl-client-stream (usocket:socket-stream sock)
                                  :hostname host)))

(connect "github.com" 443)
;; > Error: The TLS/SSL connection on handle #<A Foreign Pointer #x2363D60> has been closed (return code: 5).
;; >        SSL error queue is empty.

(connect-stream "github.com" 443)
;; #<SSL-STREAM for #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/1976) #xD01FFCE>>
````

cl+ssl tip (50a857368a5fd9d54c02ff2bcb50f406a1b81250)
CCL version 1.11-r16635  (WindowsX8632)
OpenSSL version 1.0.2l
SBCL versions 1.3.9 and 1.3.12

mtstickney:
Without this set, Windows 7 systems connecting with the default
SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 options will effectively be using the
option SSL_OP_NO_SSLv3. According to a comment in the openssl source:

> SSL_OP_NO_X disables all protocols above X if there are some protocols below X enabled.

So in that case, we're essentially telling OpenSSL to /only/ use SSLv2,
which most servers won't allow at this point.

fixes #50

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -85,7 +85,7 @@ session-resume requests) would normally be copied into the local cache before pr
 
 (defconstant +SSL-OP-ALL+ #x80000BFF)
 
-(defconstant +SSL-OP-NO-SSLv2+   #x00000000)
+(defconstant +SSL-OP-NO-SSLv2+   #x01000000)
 (defconstant +SSL-OP-NO-SSLv3+   #x02000000)
 (defconstant +SSL-OP-NO-TLSv1+   #x04000000)
 (defconstant +SSL-OP-NO-TLSv1-2+ #x08000000)


avodonosov:
Where do you look up the constant value? I'm googling for 20 minutes already.

deadtrickster:
https://github.com/openssl/openssl/blob/72257204bd2a88773461150765dfd0e0a428ee86/include/openssl/ssl.h#L397

deadtrickster:
Apparently managing different openssl versions is even more painful now, when they changed lot of constants.

avodonosov:
@deadtrickster 
Thanks for the link - even direct search in the openssl github repo didn't find anything for me.
I guess the constant was removed because new version assumes v2 is disabled by default. Unless this flag is reused for other meaning we are safe.

The fix looks good to me.

avodonosov:
@mtstickney, thanks for the report and the investigation + fix.

Issue #43, opened 2017-05-24
SBCL:The alien function "CRYPTO_num_locks" is undefined


equwal:
`The alien function "CRYPTO_num_locks" is undefined.`
non-ssl http requests work fine. I have tried completely reinstalling quicklisp, and was unfortunately unable to resolve the problem. I have the package `openssl` installed, and I am on arch linux.

The problem appears to be with the cffi bindings that are used. I have been getting this error on `drakma` and `dexador` when attempting to get ssl pages.

stassats:
Works fine here, your ssl must be broken.

equwal:
I have another computer also running arch linux. It doesn't have this problem either. I am going to try reinstalling everything I can, maybe something was corrupted.

equwal:
You were correct. I needed to install https://www.archlinux.org/packages/extra/x86_64/openssl-1.0/, in addition to the default core openssl package.

PR #59, opened 2017-12-12
Add "libssl.so.1.0.2m" to the list of paths for UNIX.

lovrolu:
Even though I had OpenSSL 1.0 installed as per https://github.com/cl-plus-ssl/cl-plus-ssl/issues/43, version 1.1 was getting loaded because "libssl.so.1.0.2m" wasn't listed in the paths.


avodonosov:
Thank you merged.

BTW, cl+ssl now allows user to load foreign libraries himself, it may be helpful if you face a similar problem in future. See example in the :CL+SSL-FOREIGN-LIBS-ALREADY-LOADED section https://common-lisp.net/project/cl-plus-ssl/.

Pull requests are also welcome.

PR #62, opened 2018-02-09
Allow libssl 1.0.2k for Unix in reload.lisp

kini:
CentOS, a common Linux distribution, currently ships OpenSSL 1.0.2k.
I also reordered the list of libssl versions in ascending order of
(possible) age.


avodonosov:
OK. Note also, cl+ssl user can control directly what native library is loaded: https://common-lisp.net/project/cl-plus-ssl/#libs-already-loaded

kini:
Thanks! I wanted to add it not for myself, but as a default for users who simply rely on cl+ssl as a dependency and who don't want to get into the details by loading the library themselves.

avodonosov:
Thank you too. Trying versions in the order of newness makes sense.

Issue #56, opened 2017-09-08
On OSX default libssl is sub-optimal?


jtkifa:
I was having problems with likely missing TLS 1.2 support until I changed
reload.lisp to have

_(cffi:define-foreign-library libssl
  (:darwin (:or 
              **"/opt/local/lib/libssl.dylib"** ;; this is new - default OSX libssl does not support TLS 1.2?
              ;; perhaps want the openssl in homebrew  here too ?
              "libssl.dylib" "/usr/lib/libssl.dylib"))_

In particular, I couldn't load https://ssd.jpl.nasa.gov/horizons.cgi in drakma

avodonosov:
I don't know what is the default libssl on OSX. 

What error message do you have when loading https://ssd.jpl.nasa.gov/horizons.cgi in drakma?

jtkifa:
The message is: 

`error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure`

and when loading one sees:

style-warning: Undefined alien: "TLSv1_1_client_method"
style-warning: Undefined alien: "TLSv1_1_server_method"
style-warning: Undefined alien: "TLSv1_1_method"
style-warning: Undefined alien: "TLSv1_2_client_method"
style-warning: Undefined alien: "TLSv1_2_server_method"
style-warning: Undefined alien: "TLSv1_2_method"

It happens when cl+ssl loads /usr/lib/libssl.dylib instead of macports version /opt/local/lib/libssl.dylib

Perhaps there could be an (environment) variable to override the library?

(cffi:define-foreign-library libssl
  (:windows (:or 
      (if (boundp 'cl::*cl+ssl-libssl*) cl::*cl+ssl-libssl*)
      "libssl32.dll" "ssleay32.dll"))
  (:darwin (:or 
              (if (boundp 'cl::*cl+ssl-libssl*) cl::*cl+ssl-libssl*)
             "/opt/local/lib/libssl.dylib"
		"libssl.dylib"
		"/usr/lib/libssl.dylib")) ...

avodonosov:
Having a variable to specify library path would be good, but I don't understand how to correctly pass it to `cffi:define-foreign-library`.

The example you provide, have you tried how it works?

[...]

PR #68, opened 2018-08-21
Add special case for loading Homebrew's OpenSSL on macOS

mneilsen:
Homebrew's "openssl" formula is keg-only:

  https://github.com/Homebrew/homebrew-core/blob/master/Formula/openssl.rb#L20

This means that Homebrew intentionally does not link OpenSSL libraries
to /usr/local/lib.  This changeset adds a special case in reload.lisp
for detecting and using Homebrew-installed OpenSSL in its default
install location, /usr/local/opt/openssl.

For more background (and motivation for specializing search paths on
macOS), see issue #56.


avodonosov:
Thank you

PR #71, opened 2019-01-16
Support loading non-RSA private keys from files

pritambaral:
install-key-and-cert uses `SSL_use_RSAPrivateKey_file`, which handles only RSA
keys. `SSL_use_PrivateKey_file` has been available in OpenSSL since [at least
0.9.8](https://github.com/openssl/openssl/blob/OpenSSL_0_9_8-stable/doc/ssl/SSL_CTX_use_certificate.pod), and has broader algorithm support (e.g., ECDSA).

This adds the `SSL_use` and `SSL_CTX_use` versions of the broader functions,
just like the RSA-specific ones at the moment, and makes install-key-and-cert
use them. This does not remove the RSA-specific functions, to be safe against inadvertently breaking external users.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -313,6 +313,17 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
     :int
   (ctx ssl-ctx)
   (type :int))
+(define-ssl-function ("SSL_use_PrivateKey_file" ssl-use-privatekey-file)
+  :int
+  (ssl ssl-pointer)
+  (str :string)
+  ;; either +ssl-filetype-pem+ or +ssl-filetype-asn1+
+  (type :int))
+(define-ssl-function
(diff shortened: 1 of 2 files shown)


avodonosov:
> SSL_use_PrivateKey_file has been available in OpenSSL since at least
0.9.8

Can you tell when exactly it was introduced? I worry about pre 0.9.8 users.

pritambaral:
> Can you tell when exactly it was introduced? I worry about pre 0.9.8 users.

I've found it in [SSLeay 0.8.1](https://github.com/openssl/openssl/blob/SSLeay_0_8_1b/doc/ssl.doc), before that was forked into OpenSSL.

avodonosov:
Thank you. Cithub says this branch has conflicts and can not be merged. Could you resolve, please?

pritambaral:
Rebased and resolved.

avodonosov:
Merged, thank you.

PR #79, opened 2019-04-19
LibreSSL is detected as 2.0 and is not compatible with OpenSSL 1.1 yet.

iarebatman:
I'm not entirely sure if this is the BEST way to fix this, but it does work on my box.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -149,6 +149,15 @@ session-resume requests) would normally be copied into the local cache before pr
 (define-crypto-function ("OpenSSL_version_num" openssl-version-num)
         :long)
 
+(define-crypto-function ("OpenSSL_version" openssl-version)
+        :string
+        '(:int))
+
+(defun libresslp ()
+    "Returns true if LibreSSL is being used"
+  (string= "LibreSSL" 
+           (subseq (openssl-version 0) 0 8)))
+
 (defun compat-openssl-version ()
   (or (ignore-errors (openssl-version-num))
       (ignore-errors (ssl-eay))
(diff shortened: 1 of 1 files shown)


avodonosov:
Thanks for tracking it down and proposing a solution. But we need a better way to detect LibreSSL. Sorry, I can't merge rhe current solution

iarebatman:
How about this instead?
I added a reference to OpenSSL_version(), which will return a string starting with "LibreSSL" or "OpenSSL".  I added a new predicate that allows you to specifically check for LibreSSL, and I replaced my previous checks based on the version numbers with the new predicate.

avodonosov:
Thank you. Meanwhile, I studied your previous approach, and we indeed can rely on version 2.0 as an indicator for LibreSSL without risk of conflict with OpenSSL in the future (the next OpenSSL will be 3.0).

I pushed a fix. Could you please test.

BTW, if you're up to contributing more, it would be good to have libressl continuous integration: https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/.travis.yml

iarebatman:
* Anton Vodonosov <notifications@github.com> [2019-04-22 00:35:03 -0700]:

> Thank you. Meanwhile, I studied your previous approach, and we indeed can rely on version 2.0 as an indicator for LibreSSL without risk of conflict with OpenSSL in the future (the next OpenSSL will be 3.0).
> 
> I pushed a fix. Could you please test.
> 
> BTW, if you're up to contributing more, it would be good to have libressl continuous integration: https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/.travis.yml
> 
Thank you for looking into this.
I can confirm that your change(s) work on my machine.
I will see what I can do about the continuous integration for LibreSSL -
I'm busy with paying work at the moment, so it may not be right away.

-- 
Regards,
Brad DeMorrow

VaclavSynacek:
I recently got the same errors with quicklisp version. Upgrading to c9957da fixed the issue also on my setup: OpenBSD 6.4 amd64, SBCL 1.4.8.openbsd.

PR #85, opened 2019-07-17
Don't load undefined libeay32 foreign-library.

mtstickney:
In commit 5860d85a8d3224b555e092b010a3bd6d7e818f31, a new foreign library named `libeay32` was defined
and added to the reload function. 1f2c1ed66e9ac273d5f71f40142249e989f70054 removed the `libeay32`
foreign-library in favor of using an alternate dll name for the `libssl`
library, but `libeay32` is still loaded in the `reload` function, which causes
an "undefined foreign library" error when `reload` is called.


avodonosov:
Thank you, merged.

avodonosov:
Would be nice to introduce a reload test to the unit tests, if you're up to contributing more, consider this.

Issue #64, opened 2018-03-17
The alien function "CRYPTO_num_locks" is undefined


whatacold:
Hi,

`dex:get` raises an undefined function error when getting a https URL,
for example(see https://github.com/fukamachi/dexador/issues/49):

    (defvar *url* "https://lispcookbook.github.io/cl-cookbook/web-scraping.html")
    (defvar *html* (dex:get *url*))

So does drakma:

    * (drakma:http-request "https://lisp.org/")
    
    debugger invoked on a SB-KERNEL::UNDEFINED-ALIEN-FUNCTION-ERROR in thread
    #<THREAD "main thread" RUNNING {1001F165F3}>:
      The alien function "CRYPTO_num_locks" is undefined.
    
    Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.
    
    restarts (invokable by number or by possibly-abbreviated name):
      0: [ABORT] Exit debugger, returning to top level.
    
    ("undefined function")
    0] 

What's shared here is that they both depend on cl+ssl, and the error is raised by cl+ssl.

I use sbcl with openssl 1.1.0g:

    # dnf list installed openssl-libs openssl
    Installed Packages
    openssl.x86_64                                                         1:1.1.0g-1.fc26                                                    @updates
    openssl-libs.x86_64                                                    1:1.1.0g-1.fc26                                                    @updates
    $ cat /proc/`pidof sbcl`/maps | grep -i cryp
    7fffefaa9000-7fffefd09000 r-xp 00000000 08:08 2649886                    /usr/lib64/libcrypto.so.1.1.0g
    7fffefd09000-7fffeff08000 ---p 00260000 08:08 2649886                    /usr/lib64/libcrypto.so.1.1.0g
    7fffeff08000-7fffeff25000 r--p 0025f000 08:08 2649886                    /usr/lib64/libcrypto.so.1.1.0g
    7fffeff25000-7fffeff31000 rw-p 0027c000 08:08 2649886                    /usr/lib64/libcrypto.so.1.1.0g

After checking the sources of openssl, I guess it's due to the changing of openssl implementation.
`CRYPTO_num_locks` is a macro in openssl v1.1.0:

    openssl-1.1.0g $ ag CRYPTO_num_locks
    CHANGES
    11112:  *) New function "CRYPTO_num_locks" that returns CRYPTO_NUM_LOCKS.
    
    include/openssl/crypto.h
[...]

PR #90, opened 2019-10-04
Un-break LibreSSL

fboundp:
Commit d3e4424 addressing issue #64 caused code paths necessary for the
correct functioning of LibreSSL to be skipped. In particular:
 - Locking callbacks are no longer set, causing crashes when CL+SSL
   is used in an application with multiple threads.
 - SSL_library_init and SSL_load_error_strings are no longer called.
 - SSL_CTX_set_tmp_rsa_callback is no longer called (probably harmless,
   as this is only relevant if one is using export grade ciphersuites).

This was an easy to make oversight. The code paths above are
predicated on COMPAT-OPENSSL-VERSION being below version 1.1, since
they are no longer necessary for versions 1.1 or later. However,
LibreSSL APIs are compatible with version OpenSSL 1.0 APIs, yet the
version number is encoded as 2.0 (technically, #x20000000).

This change updates the conditional using LIBRESSLP, so that the
necessary code paths are restored. Additionally, it adds
OPENSSL-INIT-SSL and conditionally calls it instead of
SSL-INIT-LIBRARY for OpenSSL 1.1, where OPENSSL_init_ssl supersedes
SSL_library_init and SSL_load_error_strings.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -282,6 +282,8 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
     :void)
 (define-ssl-function-ex (:vanished "1.1.0") ("SSL_library_init" ssl-library-init)
     :int)
+(define-ssl-function-ex (:since "1.1.0") ("OPENSSL_init_ssl" openssl-init-ssl)
+    :int)
 ;;
 ;; We don't refer SSLv2_client_method as the default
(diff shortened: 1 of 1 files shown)


fboundp:
Hacky workaround until this is merged:

```lisp
(when (cl+ssl::libresslp)
  (defun cl+ssl::compat-openssl-version ()
    #x10000000))
```

PR #93, opened 2019-10-20
make it work with readtable-case :invert

genworks:
Hi cl+ssl maintainers,

I hope you can accept this patch, downcasing constant names in ffi.lisp to make things work with different readtable-case settings such as :invert. 

Sorry it messes with the nice source code style of having constants in uppercase. 

Thanks,

 Dave

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -170,42 +170,42 @@ variants if you have use cases for them.)"
 (defconstant +ssl-filetype-asn1+ 2)
 (defconstant +ssl-filetype-default+ 3)
 
-(defconstant +SSL-CTRL-OPTIONS+ 32)
-(defconstant +SSL_CTRL_SET_SESS_CACHE_MODE+ 44)
-(defconstant +SSL_CTRL_MODE+ 33)
+(defconstant +ssl-ctrl-options+ 32)
+(defconstant +ssl_ctrl_set_sess_cache_mode+ 44)
+(defconstant +ssl_ctrl_mode+ 33)
 
-(defconstant +SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER+ 2)
+(defconstant +ssl_mode_accept_moving_write_buffer+ 2)
 
-(defconstant +RSA_F4+ #x10001)
+(defconstant +rsa_f4+ #x10001)
 
-(defconstant +SSL-SESS-CACHE-OFF+ #x0000
+(defconstant +ssl-sess-cache-off+ #x0000
   "No session caching for client or server takes place.")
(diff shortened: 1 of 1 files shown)


genworks:
@avodonosov, Yes, thanks, my bad. Working on lowercasing the references in the code, will push a new commit later today.

genworks:
@avodonosov Another approach would be to uppercase all constants everywhere. As long as we are consistent, it should work in any readtable-case. Would that way be preferable? (it seems so, as doing it that way also will not break any client code which is referring to the constants in uppercase, and it will retain the stylistic appeal of having constants in uppercase, making them stand out more obviously when skimming through the source code).

avodonosov:
Oh, backward compatibility, right.

Yet another way is to have duplicated definitions. Deprecated old ones for compatibility and new lower case.

Upper case constants is not lisp style, I think cl+ssl does so to follow their original C names more closely.

Anyways, what fails in the :invert case? Do we have something defined upper case and referred in lower case in the code?

avodonosov:
Because so far I thought we are consistent. Previously there were bugs, in particular, failing for the modern mode of Allego, and I thought we've fixed them all. But I see Allegro modern mode still fails: https://common-lisp.net/project/cl-test-grid/library/cl+ssl.html

So yes, leaving definitions as they are and fixing code references to match the definition case is the preferred way.

Would be also good to add such test to Travis...

avodonosov:
@genworks , I downcased those several constants which are used in the code in lowercase. They are not exported, so it's unlikely anyone dependent on them.

Left all other constants as they are.

Added an :invert case to the Travis CI build. Interestingly, the cl+ssl.test system doesn't load in this mode, because its dependency ironclad fails in this case. So for now the :invert case only applies to cl+ssl system, and for cl+ssl.test I restore the :upcase mode.

genworks:
Ok -- before I noticed your work here, I went ahead and did the opposite - upcased constants in x509.lisp to match their definition. But if you're ok lowercasing the definitions, we'll just leave it like that. Deleting my fork.

genworks:
FWIW, I just went ahead and submitted a PR for ironclad, to make it work with :invert readtable-case.

genworks:
> FWIW, I just went ahead and submitted a PR for ironclad, to make it work with :invert readtable-case.

Ok, sharplispers/ironclad already merged my PR for making ironclad work with :invert readtable-case. So with next QL release you should be able to apply :invert mode to the full cl+ssl.test.

genworks:
Hopefully you still get notified of these comments even though this PR is closed?

avodonosov:
Yes, thanks. I also subscribed to watch the ironclad pr :)

Issue #72, opened 2019-02-04
Lisp BIO broken on OpenSSL 1.1.1


avodonosov:
Turns our Lisp BIOs are completely broken when we work with OpenSSL 1.1.1 - lisps crash with a segmentation fault.

I guess that's because the BIO and BIO_METHOD structures changed memory layout (some new fields added). 

Note, OpenSSL 1.1.1 makes these structures opaque for clients - it provides functions to create them and to get / set various fields.

The task in this issue is to implement 1.1.1 support without breaking the old versions compatibility.

avodonosov:
@phmarek, most Lisp BIO is only used if you pass :unwrap-stream- p nil to make-ssl-client-stream, or on Lisps where stream-fd method is not implemented.

Otherwise file descriptor is extracted from the original lisp stream and passed to OpenSSL, so OpenSSL can perform IO itself, without user provided BIO.

Also, the breakage of Lisp BIO comes witn 1.1.1, not just 1.1

PR #97, opened 2020-05-15
Implement stream-fd for ABCL

easye:
This fixes for broken BIO on OpenSSL 1.1.1 ff. which causes ABCL to
segfault.

This hack allows ABCL to get at the underlying file descriptor using
undocumented portions of the JVM that are not guaranteed to be stable.

Tested under openjdk8 and openjdk11.

c.f. <https://github.com/cl-plus-ssl/cl-plus-ssl/issues/72>,
<https://abcl.org/trac/ticket/464>.

### DIFF
--- a/src/streams.lisp
+++ b/src/streams.lisp
@@ -208,7 +208,7 @@
   (when unwrap-stream-p
     (let ((fd (stream-fd socket)))
       (when fd
-  (setf socket fd))))
+        (setf socket fd))))
   (etypecase socket
     (integer
      (install-nonblock-flag socket)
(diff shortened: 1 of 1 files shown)


easye:
Hmm.  Somehow my git-fu messed up, and I pushed this patch on master.  Lemme know if that is a problem for accepting.

easye:
> Hmm. Somehow my git-fu messed up, and I pushed this patch on master. Lemme know if that is a problem for accepting.

I did manage to push a branch, so please use that pull request instead. 

Superseded by #98

PR #98, opened 2020-05-16
Implement stream-fd for ABCL

easye:
This fixes for broken BIO on OpenSSL 1.1.1 ff. which causes ABCL to
segfault.

This hack allows ABCL to get at the underlying file descriptor using
undocumented portions of the JVM that are not guaranteed to be stable.

Tested under openjdk8 and openjdk11.

c.f. <https://github.com/cl-plus-ssl/cl-plus-ssl/issues/72>,
<https://abcl.org/trac/ticket/464>.

### DIFF
--- a/src/streams.lisp
+++ b/src/streams.lisp
@@ -486,3 +486,24 @@ may be associated with the passphrase PASSWORD."
 #+lispworks
 (defmethod stream-fd ((stream comm::socket-stream))
   (comm:socket-stream-socket stream))
+
+#+abcl
+(progn
+  (require :abcl-contrib)
+  (require :jss)
+
+  ;;; N.b. Getting the file descriptor from a socket is not supported
+  ;;; by any published JVM API, but the following private data
+  ;;; structures have been present from openjdk6 through openjdk14 so
+  ;;; there's hope that this is somewhat unofficially official.
+  (defmethod stream-fd ((stream system::socket-stream))
+    (flet ((get-java-fields (object fields) ;; Thanks to Cyrus Harmon
+             (reduce (lambda (x y)
+                       (jss:get-java-field x y t))
+                     fields
+                     :initial-value object)))
+      (ignore-errors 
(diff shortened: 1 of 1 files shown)


easye:
Supersedes #97

avodonosov:
Hi @easye , thank you for the patch.

It might be better to wrap the stream-fd body into ignore-errors, incase in future versions of Java these fields are absent or if reflection is forbidden (in java modules environment - I suppose we access private java fields using jss:get-java-field?).

If LispBIOs are fixed in the future, then returning nil instead of signaling an error from stream-fd may let the LispBIOs to handle the interactions.

Also, this let formatting is unusual for cl+ssl code, let's have it as
```common-lisp
  (defmethod stream-fd ((stream system::socket-stream))
    ;;; Don't ask...
    (let* ((s0 (#"getWrappedInputStream" (two-way-stream-input-stream stream)))
           (s1 (jss:get-java-field s0 "in" t))
           (s2 (jss:get-java-field s1 "ch" t)))
      (jss:get-java-field s2 "fdVal" t))))
```

easye:
@avodonosov Addressed points in your review; please review again when you have a a chance.

metayan:
This change breaks both CCL and SBCL.
CCL gives:
~~~
Reader error on #<BASIC-FILE-CHARACTER-INPUT-STREAM
 ("/src/scm/lisp/cl-plus-ssl/src/streams.lisp"/5 UTF-8) #x3020021697ED>,
 near position 19991, within "fields (#\"getWrapped":
Undefined character #\" in a #\# dispatch macro.
~~~

easye:
> Neither SBCL nor CCL likes this change.
> CCL gives:
> 
> ```
> Reader error on #<BASIC-FILE-CHARACTER-INPUT-STREAM
>  ("/src/scm/lisp/cl-plus-ssl/src/streams.lisp"/5 UTF-8) #x3020021697ED>,
>  near position 19991, within "fields (#\"getWrapped":
> Undefined character #\" in a #\# dispatch macro.
> ```

Acknowledged.  Fixing.

Issue #99, opened 2020-05-24
Implementations that are not ABCL fail to load src/streams.lisp


easye:
As noted <https://github.com/cl-plus-ssl/cl-plus-ssl/pull/98#issuecomment-633098593>, the recent patch for ABCL streams unfortunately adversely effected of the usability of `master` for other implementations.

easye:
@avodonosov Apologies for breaking master.  Here's the fix.

PR #100, opened 2020-05-24
Restore loading on non-ABCL implementations

easye:
Mea culpa on the fact that the reader still reads the forms under
```cl:*features*``` conditionals regardless of their evaluation status.

FIxes <https://github.com/cl-plus-ssl/cl-plus-ssl/issues/99>


avodonosov:
@easye, thank you for the follow up.

PR #101, opened 2020-05-27
Fix automatic newline conversion to CRLF by Git

hineios:
Although the current `.gitattributes` file is intended to achieve
this, by bypassing any local definition, the current definition does not
work due to some unintended effects.

As per Git's
documentation (https://git-scm.com/docs/gitattributes#_description),
an attribute can have one of four states: Set, Unset, Set to a value,
Unspecified.
This, in addition to the documentation of the text
attribute (https://git-scm.com/docs/gitattributes#_text), dictate
that the current specification gets git to consider the text attribute
as Unspecified, where the value of `core.autocrlf` will be used.

This commit corrects the attribute text, which will force Git
to not attempt any line ending conversions on checkin or checkout.

I believe that setting the attribute text to the string value of "false" would achieve the same result, but I have not tested that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,4 +8,4 @@
 # And it general, I don't like any magic conversions. Better
 # to work with files as they are.
 
-* text=false
+* -text


avodonosov:
Thank you!

avodonosov:
@hineios , I'm reading the docs you refer, not sure I understood you right.

What is the intention of your pull request - prevent automatic conversion or enable them?

avodonosov:
@hineios , we had the text attribute prefixed with a dash, according to your first link, this means Unset, right? Why it hasn't worked?

hineios:
The intention is to prevent automatic conversion.

For example, if in Windows with `autocrlf=true`, the clonning of the repo would be done converting all line endings to CRLF (this was what led me to discover the problem).
With CRLF, SBCL isn't able to load CL+SSL because of the format in `+default-cipher-list+`.

My commit fixes this by using the "dash property name" to unset the text property.
I believe that what was happenning before was that Git was assuming the text property to be Unspecified, which in turn causes the clone behaviour to use `autocrlf`.

Why was it Unspecified? I think it is the nuance of using `false` and not `"false"`.
But since the docs recommend the usage of the dash prefix, I went with it.

avodonosov:
@hineios , I misread the diff, I though you are replacing dash with false, not the other way around. Now I understand.

Thank you.

PR #102, opened 2020-06-02
abcl: fix getting the underlying stream fd for openjdk14

easye:
@avodonosov Merge please for ABCL:  fixed openjdk14+

### DIFF
--- a/src/streams.lisp
+++ b/src/streams.lisp
@@ -493,18 +493,25 @@ may be associated with the passphrase PASSWORD."
   (require :jss)
 
   ;;; N.b. Getting the file descriptor from a socket is not supported
-  ;;; by any published JVM API, but the following private data
-  ;;; structures have been present from openjdk6 through openjdk14 so
-  ;;; there's hope that this is somewhat unofficially official.
+  ;;; by any published JVM API, so every JVM implementation may behave
+  ;;; somewhat differently.  By using the ability of
+  ;;; jss:get-java-fields to access private fields, it is usually
+  ;;; possible to "find" an access path to read the underlying integer
+  ;;; value of the file decriptor, which is all we need to pass to
+  ;;; SSL.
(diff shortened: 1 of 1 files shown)


src/streams.lisp
```diff
@@ -493,18 +493,25 @@ may be associated with the passphrase PASSWORD."
...
-  ;;; structures have been present from openjdk6 through openjdk14 so
-  ;;; there's hope that this is somewhat unofficially official.
+  ;;; by any published JVM API, so every JVM implementation may behave
+  ;;; somewhat differently.  By using the ability of
+  ;;; jss:get-java-fields to access private fields, it is usually
+  ;;; possible to "find" an access path to read the underlying integer
+  ;;; value of the file decriptor, which is all we need to pass to
+  ;;; SSL.
   (defmethod stream-fd ((stream system::socket-stream))
+    (setf *debug* stream)
```
avodonosov, line 503: I can't find the definition for the `*debug*` variable, neither in cl+ssl sources nor in CLHS.
Do you want to leave it like this or it's some forgotten piece?

easye: Leaving in the ```*debug*``` was inadvertent on my part, but it really shouldn't hurt that much to retain a reference to the last argument to ```stream-fd```.  I guess there are situations when this might not be friendly like if someone is trying to get the GC to clear such references.  

I will remove the use of ```*debug*``` to resubmit the patch when I have a chance.

easye:
superseded by #103

avodonosov:
@easye , you could keep the old pull request open, it was enough to just modify your branch and the PR would reflect the new code.

easye:
I will do that going forward with additional patches I throw yer way.  Some people like history on git branches to be immutable, a position I can dig as I use mercurial with a git/svn/perfoce bridge for everything these days.

PR #103, opened 2020-06-03
abcl: fix accessing underlying fd with openjdk14

easye:
Removed inadvertent *DEBUG* symbol.  Fix working with openjdk8 or lower.

supersedes #102

### DIFF
--- a/src/streams.lisp
+++ b/src/streams.lisp
@@ -493,18 +493,26 @@ may be associated with the passphrase PASSWORD."
   (require :jss)
 
   ;;; N.b. Getting the file descriptor from a socket is not supported
-  ;;; by any published JVM API, but the following private data
-  ;;; structures have been present from openjdk6 through openjdk14 so
-  ;;; there's hope that this is somewhat unofficially official.
+  ;;; by any published JVM API, so every JVM implementation may behave
+  ;;; somewhat differently.  By using the ability of
+  ;;; jss:get-java-fields to access private fields, it is usually
+  ;;; possible to "find" an access path to read the underlying integer
+  ;;; value of the file decriptor, which is all we need to pass to
+  ;;; SSL.
   (defmethod stream-fd ((stream system::socket-stream))
     (flet ((get-java-fields (object fields) ;; Thanks to Cyrus Harmon
              (reduce (lambda (x y)
                        (jss:get-java-field x y t))
                      fields
-                     :initial-value object)))
+                     :initial-value object))
+           (jvm-version ()
+             (read
+              (make-string-input-stream
+               (java:jstatic "getProperty" "java.lang.System"
+                                          "java.specification.version")))))
       (ignore-errors
        (get-java-fields (java:jcall "getWrappedInputStream"  ;; TODO: define this as a constant
                                     (two-way-stream-input-stream stream))
-                        '("in" "ch" "fdVal"))))))
-
-
+                        (if (< (jvm-version) 14)
+                            '("in" "ch" "fdVal")
+                            '("in" "this$0" "sc" "fd" "fd")))))))


avodonosov:
@easye , merged, thank you.

avodonosov:
@easye , all ABCL Travis builds fail now with "Class not found: com.sun.jna.Pointer". How can we fix that?

https://travis-ci.org/github/cl-plus-ssl/cl-plus-ssl/builds/694213796

avodonosov:
This started after pull request #98

easye:
> @easye , all ABCL Travis builds fails now with "Class not found: com.sun.jna.Pointer". How can we fix that?
> 
> https://travis-ci.org/github/cl-plus-ssl/cl-plus-ssl/builds/694213796

Hmm.  We need to have Apache Maven Aether libraries in the test environment to download and use the "jna.jar" artifact for ```(require :jna)``` to succeed.  

When I get some time, I'll take a look at augmenting the CI build to do this.

avodonosov:
@easye , I've tested on the pre- pull-request 98 version which was passing previously, not it fails.

So it seems PR 98 is not the cause, maybe new versions of quicklisp libraries cause this.

I've added some tracing to the Travis tests initialization: https://github.com/cl-plus-ssl/cl-plus-ssl/commit/0a7f0798abba7f323484ff9682748fbe892c898d

Looks like the error happens when we load :cffi (the output in the Travis logs is somewhat confusing - maybe this is a result of interleaving stdout and stderr).

The .travis.yaml fragment relevant to cffi loading:

```
         -e '#+abcl
             (progn
                (format t \"Loading abcl-asdf and switching maven repo URL to HTTPS (see https://github.com/armedbear/abcl/issues/151)~%\")
                (require :abcl-contrib)
                (format t \"abcl-contrib loaded...~%\")
                (require :abcl-asdf)
                (format t \"abcl-asdf loaded...~%\")
                (setf (symbol-value (read-from-string \"abcl-asdf::*default-repository*\"))
                      \"https://repo1.maven.org/maven2/\")
                (format t \"abcl-asdf::*default-repository* assigned the HTTPS URL.~%\"))'
         -e '(ql:quickload :cffi)'
```

So :abcl-contrib is loaded, maven URL is corrected to HTTPS. Why JNA is unavailable? Isn't JNA a part of abcl-contrib?

easye:
> So :abcl-contrib is loaded, maven URL is corrected to HTTPS. Why JNA is unavailable? Isn't JNA a part of abcl-contrib?

The ASDF definition for jna <https://github.com/armedbear/abcl/blob/master/contrib/mvn/jna.asd>  whose loading ```(require :jna)``` affects contains a URI reference.  At runtime, if ```jna.jar``` isn't present on the local filesystem under ```<file:~/.m2/…>```, ABCL-ASDF uses the Maven Aether libraries to download it.  

Therefore, ostensibly, the patch to access the underlying stream fd shouldn't have triggered this problem, as it should have been present previously.

avodonosov:
Yes, that patch is unrelated - I've tested a version prior the patch and it fails now (but it was passing tests just recently).

As you may remember, Maven central now rejects HTTP requests, so in cl+ssl Travis tests we change the default maven URL to HTTPS (see the .travis.yaml fragment above).

But we can do that only _after_ loading abcl-asdf. Now, if abcl-asdf tries to load JNA during its load, the JNA load is attempted before the Maven URL is fixed.

Let's suppose JNA fails to load during abcl-asdf loading. But cffi also include `(require 'jna)`, which happens after the maven URIL corrected. Should this second `(require 'jna)` load it successfully?

easye:
> As you may remember, Maven central now rejects HTTP requests, so in cl+ssl Travis tests we change the default maven URL to HTTPS (see the .travis.yaml fragment above).
> 
> But we can do that only _after_ loading abcl-asdf. Now, if abcl-asdf tries to load JNA during its load, the JNA load is attempted before the Maven URL is fixed.


Ah, subtle.  Since you are now testing abcl-1.6.1 which uses ```https``` for Maven, you shouldn't need such patching anymore.  Again, not sure why a previous successful test now fails.

avodonosov:
Actually, roswell now installs ABCL 1.7.0, a snipped from Travis logs:
```
(lisp-implementation-type): Armed Bear Common Lisp
(lisp-implementation-version): 1.7.0
```

easye:
> Let's suppose JNA fails to load during abcl-asdf loading. But cffi also include `(require 'jna)`, which happens after the maven URIL corrected. Should this second `(require 'jna)` load it successfully?

Yes, the second ```require``` should work but I think this is a bug in ABCL.  Check that the ```:jna``` symbol is present in ```cl:*features``` after the unsuccessful first invocation to confirm.

avodonosov:
BTW, the last successful build was on ABCL: 1.6.0

avodonosov:
Added `*features*` to the tracing, the :jna is ablent

```
abcl-contrib loaded...

abcl-asdf loaded...
*features*: (QUICKLISP QUICKLISP-SUPPORT-HTTPS ROS.WITHOUT-BIN ASDF3.3 ASDF3.2 ASDF3.1 ASDF3 ASDF2 ASDF OS-UNIX ROS.INIT NIO X86-64 UNIX LINUX ARMEDBEAR ABCL COMMON-LISP ANSI-CL CDR6 MOP PACKAGE-LOCAL-NICKNAMES)
```

easye:
> Added `*features*` to the tracing, the :jna is ablent

Well, that makes no sense to me currently.  

I am in the process of duplicating the cl-plus-ssl pipeline on a Linux instance with roswell to more helpfully assess what is going wrong.  But I am also still finishing release abcl-1.7.0, mainly getting the Maven artifacts published, so my time is rather fractured at this moment.  I'll let you know more when I have further information.

easye:
I believe that the java runtime being used by roswell to run abcl-1.7.0 is somehow faulty.

<https://travis-ci.org/github/armedbear/cl-plus-ssl/jobs/694930881#L190>

The ABCL CI has a script which installs a known good version of AdoptJDK <https://github.com/armedbear/abcl/blob/master/ci/install-adoptjdk.bash> which I might end up using.

I'll keep hacking away at my versions of Travis CI in <https://gitub.com/armedbear/abcl> to get a fix.

easye:
The Java version used by the cl-plus-ssl Travis is 11.0.7 <https://travis-ci.org/github/armedbear/cl-plus-ssl/jobs/694935357#L665> which should be fine.  

Back to being confused…

avodonosov:
@easye, don't be under stress, see yourself what to address first, abcl release or this issue.

If you want we can revert something (but no much sense as the issue seems to be caused not by cl+ssl changes, but by Travis env changes - abcl version or quicklisp or roswell).

BTW, this project has debug builds enabled in Travis - we can start a debug build and Travis will wait for us to connect the build machine via ssh. If you want I can assist you in getting there.

avodonosov:
I experimented locally, using ABCL 1.7.0, after (`require :jna)` finishes without errors, the `*features*` do not contain `:jna`

easye:
> I experimented locally, using ABCL 1.7.0, after (`require :jna)` finishes without errors, the `*features*` do not contain `:jna`

Uggh.  I was wrong about our CL:REQUIRE semantics.  For ABCL, the first ```(require :asdf)``` is intercepted by ASDF before the "native" require mechanism fires.  For systems loaded via ASDF we do not add a symbol to ```*features*```.   @avodonosov Should we?

easye:
<https://github.com/cl-plus-ssl/cl-plus-ssl/compare/master...armedbear:easye/abcl/20200605c> is my testing for diagnostics.  It seems Roswell just fails to invoke the JVM correctly.  Fiddling around Area 51 now… ;)

avodonosov:
@easye, interesting, the latest Travis build has passed. Maybe something was fixed on the roswell side...

PR #112, opened 2020-12-22
Upgrade to TLS1.2 or higher, and other stuff

phmarek:
Hi,

here are a some changes that you might useful. Please cherry-pick whatever you like, thanks!


avodonosov:
@phmarek , thank you, I will review a bit later

src/verify-hostname.lisp
```diff
@@ -97,10 +97,11 @@
   (when dns-names
     (error 'unable-to-match-altnames))
   ;; TODO: we are matching only first CN
```
avodonosov: so this  TODO can be removed?

src/verify-hostname.lisp
```diff
@@ -97,10 +97,11 @@
...
-  (alexandria:if-let ((cn (first (certificate-subject-common-names cert))))
-    (progn
-      (or (try-match-hostname cn hostname)
-          (error 'unable-to-match-common-name)))
+  (alexandria:if-let ((cns (certificate-subject-common-names cert)))
+    (or (some (lambda (cn)
+                (try-match-hostname cn hostname)) 
+              cns)
+      (error 'unable-to-match-common-name))
```
avodonosov: we will need to align the OR branches

src/verify-hostname.lisp
phmarek: Well, yes, my patch removes it.

src/verify-hostname.lisp
phmarek: Hmm, I can't see that misalignment?!

avodonosov:
I am in doubt about changing the min protocol version to 1.2.

To somebody this may mean a production outage. As I understand, there are no exploitable vulnerabilities known in TLS 1.1. today.

If it was insecure to allow TLS 1.1 wouldn't OpenSSL disable by default?

src/verify-hostname.lisp
avodonosov: I encountered it when reviewing a specific commit. The final version does not have it, so never-mind.

image

src/x509.lisp
```diff
@@ -195,12 +195,16 @@ we are going to pass them to CFFI:WITH-POINTER-TO-VECTOR-DATA)"))
...
-                        (alexandria:if-let ((string (try-get-asn1-string-data data '(#.+v-asn1-iastring+))))
-                          string
-                          (error "Malformed certificate: possibly NULL in dns-alt-name"))))))
+          (flet ((alt-name-to-string (alt-name)
+                   (cffi:with-foreign-slots ((type data) alt-name (:struct general-name))
+                     (case type
+                       (#. +GEN-IPADD+
+                         (let ((address (asn1-string-bytes-vector data)))
+                           (usocket:host-to-hostname address)))
+                       (#. +GEN-DNS+
```
avodonosov, line 204: Why do you want read-time value here?

I think to remove it.

phmarek:
Well, see https://security.stackexchange.com/questions/57878/what-are-the-main-vulnerabilities-of-tls-v1-1 - it's about bad (as in "now defined as insecure") algorithms, like MD5.

TLS 1.2 is from 2008.

If you search for "deprecate tls1.1" you will find lots of results - eg. https://tools.ietf.org/id/draft-ietf-tls-oldversions-deprecate-02.html

I believe that we should go *at least* to 1.2; security of the masses should take precedence over laziness of the individuals ;)

src/x509.lisp
phmarek: `CASE` doesn't evaluate the keys, and we need the numeric value of the constants here.
(http://clhs.lisp.se/Body/m_case_.htm)

avodonosov:
@phmarek , I have pulled your changes but removed the  default value for `min-proto-version`. See the commit following yours.

My previous comment about no known exploitable vulnerabilities in TLS 1.0 and 1.1 is based on the first link from your reply, actually.

I am very opposed to any destructive changes in CL ecosystem, some breakages can take years to repair, some will never be repaired. It's not about laziness, there may be no maintainer at all. And even for active maintainers, better to save their time as much as possible.

Even Java does not block TLS 1.0 and 1.1 by default. And it is maintained by a commercial corporation Oracle, and used by many enterprise systems. Let's not be "more Catholic than the Pope".

On application level people are migrating actively now, but it's not the time for a library to force this yet.

avodonosov:
What is pending: update the doc, export the constants for TLS versions from the package. But I am in doubt what character case to use for public constants. In the ffi.lisp they are declared in up case: `+TLS1-2-VERSION+`

But Slime completion downcases symbols, so it's very likely for users to end up with `cl+ssl:+tls1-2-version+` in their code, which will break on non-classic reader mode (Allegro modern mode, for example).

Even the current cl+ssl code suffers from that - the package.lisp refers some constants as lower case, so we export different, unbound symbols from the package in :preserve or :invert readtable-cases.

avodonosov:
@phmarek thank you for your contributions.

Can you remind me about this code version you had in the past: https://github.com/phmarek/cl-plus-ssl/blob/3566522e07f0a5c234bb1dee16e7d0b218e1c419/src/ffi.lisp#L194

I consider adopting your `eval` solution, it is simpler than the current one (https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/src/ffi.lisp#L40). However, your comment says it "breaks things". Breaks how?

phmarek:
Hi Anton,

> I have pulled your changes

Thanks a lot!


> but removed the  default value
> for `min-proto-version`. See the commit following yours.
> 
> My previous comment about no known exploitable vulnerabilities in TLS
> 1.0 and 1.1 is based on the first link from your reply, actually.

Yes, I'm well aware of that.

But if MITM can downgrade your SSL connection to TLS1.1 or less,
the connection is no longer secure!

If you're looking for reasons to diable TLS<1.2, see here:

https://security.stackexchange.com/questions/223603/main-reasons-to-disable-tls-1-1-1-2


> I am very opposed to any destructive changes in CL ecosystem, some
> breakages can take years to repair, some will never be repaired. It's
> not about laziness, there may be no maintainer at all. And even for
> active maintainers, better to save their time as much as possible.

I understand you.

Still, IMO it's better to point out some problem (using insecure 
defaults)
by breaking something than keeping them insecure forever!


> Even Java does not block TLS 1.0 and 1.1 by default. And it is
> maintained by a commercial corporation Oracle, and used by many
> enterprise systems. Let's not be "more Catholic than the Pope".

Microsoft disables TLS<1.2:
https://blogs.windows.com/msedgedev/2020/03/31/tls-1-0-tls-1-1-schedule-update-edge-ie11/

OpenJDK disables TLS<1.2
https://bugs.openjdk.java.net/browse/JDK-8214443
(approved in 2018!)

The German BSI deprecated TLS<1.2:
https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-2.pdf?__blob=publicationFile&v=10
(Page 6, 3.2: SSL/TLS versions)


> On application level people are migrating actively now, but it's not
> the time for a library to force this yet.

Well, TLS1.2 is available since 2008, so it's time to _force_ migration, 
IMO.

phmarek:
> What is pending: update the doc,

I'll try to prepare something.


> export the constants for TLS versions
> from the package. But I am in doubt what character case to use for
> public constants. In the ffi.lisp they are declared in up case:
> `+TLS1-2-VERSION+`
> 
> But Slime completion downcases symbols, so it's very likely for users
> to end up with `cl+ssl:+tls1-2-version+` in their code, which will
> break on non-classic reader mode (Allegro modern mode, for example).
> 
> Even the current cl+ssl code suffers from that - the package.lisp
> refers some constants as lower case, so we export different, unbound
> symbols from the package in :preserve or :invert readtable-cases.

Yeah, I guess we should make the symbols lowercase.

phmarek:
> Can you remind me about this code version you had in the past:
> https://github.com/phmarek/cl-plus-ssl/blob/3566522e07f0a5c234bb1dee16e7d0b218e1c419/src/ffi.lisp#L194
> 
> I consider adopting your `eval` solution, it is simpler than the
> current one
> (https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/src/ffi.lisp#L40).
> However, your comment says it "breaks things". Breaks how?

Uh, it's been some time.

IIRC I had problems with the function signature (including return type!) 
being
kept in some way - so calling such a (redefined) function would give
`Function declared as returning nothing returned (1 2 3)`
or some similar error messages during runtime.

avodonosov:
@phmarek , OpenJDK does not disable TLS < 1.2. This ticket is about removing the TLS 1.0 and 1.1 from the requirements for conforming implementation - if anyone produces Java without TLS 1.1 it will be considered conformant.

The recent Java 15 comes with TLS 1.0 enabled.

phmarek:
> OpenJDK does not disable TLS < 1.2.

Sorry, perhaps I pasted the wrong link. 
https://bugs.openjdk.java.net/browse/JDK-8202343?subTaskView=unresolved "Disable TLS 1.0 and 1.1" links to a changeset (https://github.com/openjdk/jdk/commit/3a4b90f0) referencing jdk-16 and jdk-17; but the bug report also lists backports for Java 8, 7, and Oracle Java 11.

How about defaulting to a special that can be set once per image (eg. via `.sbclrc`), like in 42972fa8497a78e5bd58422dcec78bbf4e8ac290?

avodonosov:
@phmarek , that's an interesting ticket. But right now the latest JDK (I downloaded a day ago), has the following in conf/security/java.security

```
jdk.tls.disabledAlgorithms=SSLv3, RC4, DES, MD5withRSA, DH keySize < 1024, \
    EC keySize < 224, 3DES_EDE_CBC, anon, NULL, \
    include jdk.disabled.namedCurves
```

Also, I think OpenSSL users can disable undesirable algorithms in global OpenSSL config files ( /etc/ssl/openssl.cnf or /usr/lib/ssl/openssl.cnf, or some other - I don't know which one). 

It's interesting why OpenSSL does not disable the unsafe versions by default in new contexts. For example, they disable SSL v2: https://www.openssl.org/docs/man1.0.2/man3/SSLv3_method.html
> The SSL_OP_NO_SSLv2 option is set by default, and would need to be cleared via SSL_CTX_clear_options(3) in order to enable negotiation of SSLv2.

So, I think there is no urgency in disabling TLS 1 and 1.1

avodonosov:
Cherry picked your commit adding two missing constants.

Do you see any use from a global var if it's initialized to nil?

phmarek:
> Cherry picked your commit adding two missing constants.

Ack, thanks!


> Do you see any use from a global var if it's initialized to nil?

Yes -- at least there's a _central_ place to set the preferred
minimum protocol; there's no need to look through the sources
where the argument should be added.


Do you happen to know whether that context is process-global,
ie. is used for incoming as well as outgoing connections?
(I'm thinking of eg. DRAKMA here.)

That would be important.

avodonosov:
Drakma's code is here: https://github.com/edicl/drakma/blame/master/util.lisp#L330

phmarek:
Yeah, thanks.

This uses the global context, so we just need to make sure that one uses the special variable, right?
Is it possible that the context is dumped with an image and the reused when restarting the image (potentially across incompatible OpenSSL versions?) Can we rebuild the context on every image load?

PR #113, opened 2021-01-03
Adds not-before and not-after date accessors for x509 certs

mtm:
Adds:

`get-not-after-time`
`get-not-before-time`
`decode-asn1-utctime`

as well as some tests


test/package.lisp
```diff
@@ -4,7 +4,7 @@
 
 (defpackage :cl+ssl.test
   (:use :cl
-        :5am))
+        :fiveam))
```
avodonosov, line 7: Why?

mtm: ah, sorry about that.  local gorp, reverted

src/x509.lisp
```diff
@@ -234,3 +253,15 @@ we are going to pass them to CFFI:WITH-POINTER-TO-VECTOR-DATA)"))
         if cn collect cn
         if (not cn) do
            (loop-finish)))))
+
+(defun get-not-after-time (cert)
```
avodonosov, line 257: A docstrings is desirable for these new functions, saying at least about the return type - universal time.
And the same info in the index.html

mtm: docs added

mtm:
hmm, supporting older versions (pre 1.1.0) is slightly tricky: the `X509_get_notAfter` is implemented as a pre-processor macro in "x509.h".  working on it...

src/x509.lisp
```diff
@@ -234,3 +253,25 @@ we are going to pass them to CFFI:WITH-POINTER-TO-VECTOR-DATA)"))
...
+      (error "X509_get0_notAfter returned NULL"))
+    (decode-asn1-utctime asn1-time)))
+
+(defun get-not-before-time (certificate)
+  "Returns a universal-time representing the time before
+which the certificate is not valid."
+  ;; The low-level FFI changed in version 1.1.0 of OpenSSL, we'll
+  ;; try the newer one first
+  (let ((asn1-time (or (ignore-errors (x509-get0-not-before certificate))
+                       (ignore-errors (x509-get-not-before certificate)))))
```
avodonosov, line 274: Better to use conditions like `(openssl-is-at-least 1 1)` and `(libresslp)`. See for example the definition of`sk-general-name-value`.

avodonosov:
@mtm , thank you for the contribution.

We could even start with the implementation for 1.1.0+  only. Older version could be supported in future patches, maybe by the users of the older OpenSSL. Every passing day makes this need smaller and smaller.  My only requirement is to not break existing code, e.g. keep cl+ssl loadable with old OpenSSL versions.

Although I see you have almost completed cross-version support.

My biggest hesitation now is naming. In my opinion the word `get-` is redundant, shorter functional style names like `not-before-time`, `not-after-time` have the same readability. Another concern is that the already existing public function `certificate-subject-common-names` is prefixed with `certificate-` so maybe for consistency it's better to add this for the new functions as wells: `certificate-not-before-time`, `certificate-not-after-time`.

What do you think?

I admit real user code may often look like
```common-lisp
(let ((cert ...))
  (certificate-not-before-time cert))
```
So the "cert[ificate]" is mentioned both in function name and parameter, thus making the code a bit redundant / tautological.

However, it's may be easier to explore the package content with such naming and again, consistency with the existing function.

src/x509.lisp
mtm: done

mtm:
agreed on the renaming and 1.1.0+ only for now.  Supporting older versions will require some fiddly, multi-level deep  `defcstruct` work.

mtm:
Had to exclude `libressl` for now, which is a shame: libressl 2.8.3+ works but I don't see a clean way to determine its version.

Any ideas?

test/validity-dates.lisp
```diff
@@ -8,7 +8,8 @@
 (in-suite :cl+ssl.validity-dates)
 
 (test validity-dates-google-cert
-  (when (cl+ssl::openssl-is-at-least 1 1 0)
+  (when (or (cl+ssl::openssl-is-at-least 1 1 0)
+            (cl+ssl::libresslp))
```
avodonosov, line 12: (not (libresslp)) , right?

avodonosov:
For libressl version, based on its doc page the only way I see is to call 
`OpenSSL_version(OPENSSL_VERSION)` ano parse the returned version string.
http://man.openbsd.org/OPENSSL_VERSION_NUMBER.3

Or better `SSLeay_version(SSLEAY_VERSION)` that exists in old versions as well.

We don't yet have code examples discriminating by libressl version.

I am not sure, but maybe such code whould cause refactoring of the current `:sincè and `:vanished` notation. Or maybe not.

test/validity-dates.lisp
mtm: Oops, yes, of course.  fixed

PR #115, opened 2021-01-31
use versioned dylib path for libcrypto on macOS

### LINKED ISSUE #114: Mac M1: loading libcrypto in an unsafe way
When trying to install `cl+ssl` on a new M1 mac using quicklisp, it hangs and is pushing me into LDB, showing this error:

```
This is SBCL 2.0.11, an implementation of ANSI Common Lisp.
More information about SBCL is available at <http://www.sbcl.org/>.

SBCL is free software, provided as is, with absolutely no warranty.
It is mostly in the public domain; some portions are provided under
BSD-style licenses.  See the CREDITS and COPYING files in the
distribution for more information.
* (ql:quickload :cl+ssl)
To load "cl+ssl":
  Load 1 ASDF system:
    cl+ssl
; Loading "cl+ssl"
.WARNING: /usr/local/Cellar/sbcl/2.0.11/libexec/bin/sbcl is loading libcrypto in an unsafe way
fatal error encountered in SBCL pid 55608 pthread 0x101042e00:
SIGABRT received.

Welcome to LDB, a low-level debugger for the Lisp runtime environment.
ldb> 
```

From #lisp on freenode, @phoe pointed this out while helping me:
 - https://github.com/cl-plus-ssl/cl-plus-ssl/blob/ff4634a2addedc033394e93a402523532e9b470a/src/reload.lisp#L44-L49
 - https://github.com/orthecreedence/cl-async/issues/181

It seems like a solution is to add a specific version of libcrypto.

fjl:
In macOS 11, libcrypto.dylib provided by Apple will crash the process
when loading it, saying that it is being loaded in an "unsafe way". Loading
a specific version of the library explicitly is fine, and does not cause
this issue.

Fixes #114

### DIFF
--- a/src/reload.lisp
+++ b/src/reload.lisp
@@ -46,10 +46,9 @@
                 "/usr/local/opt/openssl/lib/libcrypto.dylib" ;; Homebrew
                 "/usr/local/lib/libcrypto.dylib" ;; personalized install
                 ;; System-provided libraries. Please note that in macOS >= 11.0,
-                ;; these paths may not exist in the file system anymore, but
-                ;; trying to load them via dlopen will work. This is because
-                ;; macOS ships all system-provided libraries as a
-                ;; single 'dyld cache' bundle.
+                ;; these paths may not exist in the file system anymore, but trying
+                ;; to load them via dlopen will work. This is because macOS ships
+                ;; all system-provided libraries as a single dyld_shared_cache bundle.
                 "/usr/lib/libcrypto.44.dylib"
                 "/usr/lib/libcrypto.42.dylib"
                 "/usr/lib/libcrypto.41.dylib"
--- a/src/reload.lisp
+++ b/src/reload.lisp
@@ -45,8 +45,11 @@
                 "/sw/lib/libcrypto.dylib"        ;; Fink
                 "/usr/local/opt/openssl/lib/libcrypto.dylib" ;; Homebrew
                 "/usr/local/lib/libcrypto.dylib" ;; personalized install
-                "libcrypto.dylib"                ;; default system libcrypto, which may have insufficient crypto
-                "/usr/lib/libcrypto.dylib"))
+                ;; system-provided libraries
+                "/usr/lib/libcrypto.44.dylib"
+                "/usr/lib/libcrypto.42.dylib"
+                "/usr/lib/libcrypto.41.dylib"
+                "/usr/lib/libcrypto.35.dylib"))
   (:cygwin (:or "cygcrypto-1.1.dll" "cygcrypto-1.0.0.dll")))
 
 (cffi:define-foreign-library libssl
@@ -62,8 +65,11 @@
                 "/sw/lib/libssl.dylib"        ;; Fink
                 "/usr/local/opt/openssl/lib/libssl.dylib" ;; Homebrew
                 "/usr/local/lib/libssl.dylib" ;; personalized install
-                "libssl.dylib"                ;; default system libssl, which may have insufficient crypto
-                "/usr/lib/libssl.dylib"))
+                ;; system-provided libraries
+                "/usr/lib/libssl.46.dylib"
+                "/usr/lib/libssl.44.dylib"
+                "/usr/lib/libssl.43.dylib"
+                "/usr/lib/libssl.35.dylib"))
   (:solaris (:or "/lib/64/libssl.so"
                  "libssl.so.0.9.8" "libssl.so" "libssl.so.4"))
   ;; Unlike some other systems, OpenBSD linker,
--- a/src/reload.lisp
+++ b/src/reload.lisp
@@ -45,7 +45,11 @@
                 "/sw/lib/libcrypto.dylib"        ;; Fink
                 "/usr/local/opt/openssl/lib/libcrypto.dylib" ;; Homebrew
                 "/usr/local/lib/libcrypto.dylib" ;; personalized install
-                ;; system-provided libraries
+                ;; System-provided libraries. Please note that in macOS >= 11.0,
+                ;; these paths may not exist in the file system anymore, but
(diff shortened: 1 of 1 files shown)


src/reload.lisp
```diff
@@ -45,8 +45,11 @@
                 "/sw/lib/libcrypto.dylib"        ;; Fink
                 "/usr/local/opt/openssl/lib/libcrypto.dylib" ;; Homebrew
                 "/usr/local/lib/libcrypto.dylib" ;; personalized install
-                "libcrypto.dylib"                ;; default system libcrypto, which may have insufficient crypto
-                "/usr/lib/libcrypto.dylib"))
+                ;; system-provided libraries
+                "/usr/lib/libcrypto.44.dylib"
+                "/usr/lib/libcrypto.42.dylib"
+                "/usr/lib/libcrypto.41.dylib"
+                "/usr/lib/libcrypto.35.dylib"))
```
avodonosov, line 52: Do we want to remove both "libcrypto.dylib" and /usr/lib/libcrypto.dylib"? Maybe just put them at the end of the list?

Also, do we want the MacPorts, Fink and Homebrew paths to be tried before the new versioned paths? (CFFI tries the paths in the specified order)

fjl: It depends. There is no problem with trying those other paths first, and it might help people who want to install their own version of openssl. 

The issue is mostly that loading "/usr/lib/libcrypto.dylib" or just "libcrypto.dylib" causes the crash, so that one shouldn't be in the list. It seems that Apple decided it's a bad idea to load it like that in macOS 10.15. They made it print a warning in macOS 10.15, but I guess nobody noticed because of homebrew. In macOS 11.0, they made this warning fatal, and I only managed to trigger it because homebrew has moved their default prefix to`/opt/homeberw` for arm64.

fjl: We could consider adding "/usr/lib/libcrypto.dylib" back for compatibility with very old macOS systems.

avodonosov:
I forgot about this pull request, sorry.
Do you know any simple way to check MacOS version, so that we could load the unversioned libcrypto only on old MacOSes?

fjl:
I don't know an easy/portable way to check that. But it's also futile to try, because macOS is always moving forward. Apple will not revert their decision here, and does not support old OS versions indefinitely. So if you made a change now that makes it work on older versions of macOS, the fix will become irrelevant over time.

If you have an older version of macOS, you could check which libcrypto API versions are available and then add those. I just added the versions available on macOS 11.0 Big Sur here. I'm sure these will stick around for quite a while because by removing them, Apply would break compatibility with .app binaries linked to them.

src/reload.lisp
smashedtoatoms: Any chance we can add `/opt/homebrew/opt/openssl/lib ;; Homebrew Arm64` to this list?  That would take care of us darwin arm64 people.

smashedtoatoms: Sorry, It was late last night when I made that comment.  I should have just submitted a PR.  I submitted https://github.com/cl-plus-ssl/cl-plus-ssl/pull/119 to address it.

src/reload.lisp
```diff
@@ -45,8 +45,17 @@
                 "/sw/lib/libcrypto.dylib"        ;; Fink
                 "/usr/local/opt/openssl/lib/libcrypto.dylib" ;; Homebrew
                 "/usr/local/lib/libcrypto.dylib" ;; personalized install
-                "libcrypto.dylib"                ;; default system libcrypto, which may have insufficient crypto
-                "/usr/lib/libcrypto.dylib"))
+                ;; System-provided libraries. These must be loaded with an explicit
+                ;; API version, or else the process will crash.
+                ;;
+                ;; Please note that in macOS >= 11.0, these paths may not exist in the
```
avodonosov, line 51: Why do you say "these paths"? All of them may not exists, or only the lower versions like 35? Can the version 44 not exist?

avodonosov:
@fjl, I committed the versioned paths, but keep the unversioned ones in the end of the list for now, in order to support backward compatibility. We will need to delete them eventually. See the commit. I haven't find very clear statistics of macOS 11 adoption, but it seems older versions are still actively used too.

BTW, can you tell, what of the version numbers you added correspond to what macOS version? (Are your versions only for macOS 11 or some of them work also for older macOSes?)

Another question: why does libcrypto need versioned paths, but libssl does not?

src/reload.lisp
fjl: None of these paths exist in the filesystem on my machine, which is running macOS 11.2.2 at this time. I try to explain it in the comment here because someone might have issues with loading the library later, and then that person would go looking for those files and be surprised that they are not present at all. At least that's what happened to me.

fjl:
I just added all versions which are present in the dyld_shared_cache on my machine, which is running macOS 11.2. It's very likely that some of these API versions also exist in previous versions of macOS, but I cannot verify this.

> Another question: why does libcrypto need versioned paths, but libssl does not?

I'm not sure about this. Did you test it? If Apple wants developers to not link openssl without an API version number, I think the right thing to do is using a version number for libssl, too.

src/reload.lisp
fjl: You can read more about this in the macOS Big Sur release notes: https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes

(search for dynamic linker)

> New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

avodonosov:
> I'm not sure about this. Did you test it?

I do not have MacOS.

PR #116, opened 2021-02-15
Add :canary to cffi:define-foreign-library

daewok:
See
https://common-lisp.net/project/cffi/manual/cffi-manual.html#define_002dforeign_002dlibrary

This is CFFI's more general solution to pushing
`:cl+ssl-foreign-libs-already-loaded` to `*features*`. Supporting this is nice
as it requires zero work from the user.

I tried to choose symbols that appear to be present in all versions of
OpenSSL. If one of these symbols ever disappears, the canary should be updated.

### DIFF
--- a/src/reload.lisp
+++ b/src/reload.lisp
@@ -36,7 +36,7 @@
 ;;
 ;; These are 32-bit only.
 
-(cffi:define-foreign-library libcrypto
+(cffi:define-foreign-library (libcrypto :canary "BIO_new")
   (:windows (:or #+(and windows x86-64) "libcrypto-1_1-x64.dll"
                  #+(and windows x86) "libcrypto-1_1.dll"
                  "libeay32.dll"))
@@ -75,7 +75,7 @@
                 "/usr/lib/libcrypto.dylib"))
   (:cygwin (:or "cygcrypto-1.1.dll" "cygcrypto-1.0.0.dll")))
 
-(cffi:define-foreign-library libssl
+(cffi:define-foreign-library (libssl :canary "SSL_connect")
   (:windows (:or #+(and windows x86-64) "libssl-1_1-x64.dll"
                  #+(and windows x86) "libssl-1_1.dll"
                  "libssl32.dll"


avodonosov:
Hi, thank you for the proposal.

But isn't that abuse, the docs say it's for statically linked lib?

What drawbacks we can face with this solution?

daewok:
No, I don't think it's abusing it. CFFI is just assuming that the most likely reason a symbol is already present is because the lib was statically linked into the runtime. I actually do this with SBCL and openssl so when I distribute my program I don't have to worry about a compatible openssl being installed on the target computer.

As far as I can tell, the end effect of this is nearly identical to pushing `:cl+ssl-foreign-libs-already-loaded`. The only difference is that you can't turn it off. If the canary is present at the first call to `use-foreign-library` for that lib, all future calls will turn into noops.

avodonosov:
I think this solution may break ACL and ABCL, because the canary symbols will be found in boringssl / libressl those lisps include, see https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/src/ffi.lisp#L127

Can not test it now because or Travis CI is broken currently, https://github.com/roswell/roswell/issues/463

Is it so inconvenient to put `:cl+ssl-foreign-libs-already-loaded`?

daewok:
Wow. That is gross. I think you're right that the canary would break them.

It's mildly inconvenient to add the feature, but I can manage. I thought that the canary would Just Work across all setups, but I guess not.

If you don't mind I'm going to leave this open until CI is fixed, just in case our intuitions are wrong.

avodonosov:
@daewok , CI is fixed, you can try re-running your PR on Travis CI.
Probably that will require rebasing your branch onto the master (if Travis CI's pull request testing is simply running the branch, as opposed to simulating merge and then testing).

avodonosov:
Yes, rebase is needed, I've tried to run the pull request's build and it fails as before currently.

daewok:
So the tests passed. But after looking at them, I see they unconditionally load libssl before loading cl+ssl. So there's not even a chance to expose a bug due to adding the canary on ACL and ABCL. I'm highly confident that if the right thing to do on those implementations is to load libssl even if they already have some SSL symbols present, then the canary will break things. So I'll just close this.

PR #117, opened 2021-02-22
SNI server-side support

landakram:
Add support for handling SNI on the server. The ffi functions for working with the
SNI callback have been added, and a default callback has been defined. If `:sni t` is
passed to `make-context`, the default callback is registered. The callback can be
manually specified when calling `make-context` by using the `:sni-callback`
key.

When calling `make-ssl-server-stream`, an alist for SNI configuration can be passed
using the `:sni-config` key. The alist has the following structure:

```lisp
  ((server-name-str . ((:certificate certificate-path
                        :key key-path
                        :passphrase passphrase-str)))))
```

If a servername is passed in the TLS handshake and the default callback has been
registered, the servername will be looked up in the sni-config. If a value is
present, the certificate and key for the SNI entry are set on the SSL object. If
no value is present, SSL_TLSEXT_ERR_NOACK is returned and the server will continue
with the :certificate and :key passed into `make-ssl-server-stream`.

### DIFF
--- a/src/context.lisp
+++ b/src/context.lisp
@@ -96,6 +96,8 @@
                           (verify-mode +ssl-verify-peer+)
                           (verify-callback nil verify-callback-supplied-p)
                           (cipher-list +default-cipher-list+)
+                          (sni nil)
+                          (sni-callback 'sni-callback)
                           (pem-password-callback 'pem-password-callback))
   (ensure-initialized)
   (let ((ctx (ssl-ctx-new (if method-supplied-p
@@ -134,6 +136,8 @@
                                                       (cffi:callback verify-peer-callback)
                                                       (cffi:null-pointer)))))
       (ssl-ctx-set-cipher-list ctx cipher-list)
+      (when (and sni sni-callback)
+        (ssl-ctx-set-tlsext-servername-callback ctx (cffi:get-callback sni-callback)))
       (ssl-ctx-set-default-passwd-cb ctx (cffi:get-callback pem-password-callback))
       ctx)))
 
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -233,6 +233,14 @@ session-resume requests) would normally be copied into the local cache before pr
 (defconstant +SSL-CTRL-SET-MIN-PROTO-VERSION+ 123)
 (defconstant +SSL-CTRL-SET-MAX-PROTO-VERSION+ 124)
 
+(defconstant +SSL-TLSEXT-ERR-OK+ 0)
+(defconstant +SSL-TLSEXT-ERR-ALERT-WARNING+ 1)
+(defconstant +SSL-TLSEXT-ERR-ALERT-FATAL+ 2)
+(defconstant +SSL-TLSEXT-ERR-NOACK+ 3)
+
+(defconstant +SSL-CTRL-SET-TLSEXT-SERVERNAME-CB+ 53)
+(defconstant +SSL-CTRL-SET-TLSEXT-SERVERNAME-ARG+ 54)
+
 (defconstant +SSL3-VERSION+ #x0300)
 (defconstant +TLS1-VERSION+ #x0301)
 (defconstant +TLS1-1-VERSION+ #x0302)
@@ -478,12 +486,24 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
   (larg :unsigned-long)
   (parg :pointer))
 
+(define-ssl-function ("SSL_CTX_callback_ctrl" ssl-ctx-callback-ctrl)
+  :long
+  (ctx ssl-ctx)
+  (cmd :int)
+  (fp :pointer))
+
 (define-ssl-function ("SSL_ctrl" ssl-ctrl)
     :long
   (ssl :pointer)
   (cmd :int)
   (larg :long)
   (parg :pointer))
+
+(define-ssl-function ("SSL_callback_ctrl" ssl-callback-ctrl)
+  :long
+  (ssl :pointer)
+  (cmd :int)
+  (fp :pointer))
 
 #+new-openssl
 (define-ssl-function ("SSL_CTX_set_options" ssl-ctx-set-options)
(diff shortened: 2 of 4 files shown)


landakram:
Hi there, I put together this changeset to add server-side support for SNI. I have lightly, manually tested it and verified that it works as expected. 

That is, with something like the following code:

```lisp
(let* ((tls-context (make-context :sni t))
       (sni-config (("foo.bar" (:certificate . "/tmp/foo.bar/cert.pem")
                               (:key . "/tmp/foo.bar/key.pem")
                               (:passphrase . ""))
                    ("baz.quux" (:certificate . "/tmp/baz.quux/cert.pem")
                                (:key . "/tmp/baz.quux/key.pem")
                                (:passphrase . ""))))
       (make-ssl-server-stream stream :sni-config sni-config))
  #|Do more stuff...#|)
```
I can run:

```sh
openssl s_client -crlf -tls1_2 -servername "foo.bar" -connect localhost:1965
```

and successfully connect. I can also run:

```sh
openssl s_client -crlf -tls1_2 -servername "not.handled" -connect localhost:1965
```

and see a connection error, as expected.

But I'm sure that I'm missing something... and I haven't thought at all about automated testing or docs. I'd be happy to put more effort into testing and docs, but I wanted to see whether this is a reasonable direction so far, as I have made some somewhat opinionated decisions.

After looking into some implementations, I saw SNI done two ways:

1. Keep a separate SSL_CTX for each certificate. Set the SNI callback on every SSL_CTX. In the SNI callback, select the correct SSL_CTX and set it in the SSL struct. The SSL_CTX structs hold the certificate and keys for each certificate respectively.
2. Keep a single SSL_CTX, with the SNI callback set. In the SNI callback, set the certificate and key on the SSL struct rather than SSL_CTX.

The structure of the code made option 2 more appealing to me so I went that direction. Managing multiple contexts would also work, but seemed like a bigger change and felt awkward since the code is built around a single context (the binding of `*ssl-global-context*`) at a given time.

Let me know what you think!

avodonosov:
@landakram , thank you for the contribution. I did one approach to the review, but need more time to understand it better.

avodonosov:
@landakram, your solution looks mostly fine, but I think of some simplification.

I feel the sni config is more natural to be related to context directly.

Making the main use case where context's callback uses a dynamic var, that is hold bound for the duration of ssl-accept - a little to much indirection for people to understand. Also this will not support async IO mode (that's not the primary usage scenario of cl+ssl, but we have functions for that - `nonblocking-ssl-funcall`, `input-wait`, `output-wait`).

So I am waiting in how to factorize suitable building blocks from your solution.

avodonosov:
@landakram , BTW, how do you think, when we call install-key-and-cert , does it always mean reading the disk and decrypting the key? So such a server re-reads the cert from disc for every new client socket?

avodonosov:
What also raises doubts is 

> It is also recommended that applications utilize a ClientHello callback and not use a servername callback, in order to avoid unexpected behavior that occurs due to the relative order of processing between things like session resumption and the historical servername callback.

https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_client_hello_cb.html

At the same time, 

> However, even where the ClientHello callback is used, the servername callback is still necessary in order to acknowledge the servername requested by the client.

https://www.openssl.org/docs/man1.1.1/man3/SSL_get_servername_type.html

landakram:
> BTW, how do you think, when we call install-key-and-cert , does it always mean reading the disk and decrypting the key? So such a server re-reads the cert from disc for every new client socket?

Yes, that looks to be true. It's the same situation as `make-ssl-server-stream` as far as I can tell.

> I feel the sni config is more natural to be related to context directly.

I totally agree. `make-context` returns the ssl-ctx directly, so I didn't know where to store it... I could store it on the context but I wanted to avoid doing more c interop. Maybe I could store it in a separate alist/plist/hash and remove the entry on `ssl-ctx-free`? 

> Making the main use case where context's callback uses a dynamic var, that is hold bound for the duration of ssl-accept - a little to much indirection for people to understand. 

I did this because to avoid passing the config to C... I saw a similar trick with `*pem-password*`. I'll ponder other options though as I agree that it's indirect. 

> Also this will not support async IO mode (that's not the primary usage scenario of cl+ssl, but we have functions for that - nonblocking-ssl-funcall, input-wait, output-wait).

Is this because of the use of dynamic variables or something else?

landakram:
Thinking about it more today, using a separate mapping from ssl-ctx to sni-config lets us move sni-config to `make-context` and remove the dynamic var... it that seems like a reasonable direction to you, I can look at implementing. Open to other suggestions too :stuck_out_tongue:

avodonosov:
> Is this because of the use of dynamic variables or something else?

Yes, I meant if we do

```common-lisp
  (let ((*sni-config*))
      (some-non-blocking-call))
```
The actual IO may be performed after the `let` has existed so the binding will not be in effect.

As for avoiding a dynamic var, I was thinking in directions that I now realize will not work or will not be very convent.
For example, cl+ssl provides a helper function that taks sni-config as a parameter:

```commons-lisp
(defun sni-config-handler (ssl-pointer al arg sni-config)
  ...)
```
And then users define their callback, utilizing this helper:

```common-lisp

(cffi:defcallback my-sni-callback :int ((ssl ssl-pointer) (al :int) (arg :pointer))
  (sni-config-handler ssl-pointer a1 arg '(("foo.bar" (:certificate . "/tmp/foo.bar/cert.pem")
                                            (:key . "/tmp/foo.bar/key.pem")
                                            (:passphrase . ""))
                                           ("baz.quux" (:certificate . "/tmp/baz.quux/cert.pem")
                                            (:key . "/tmp/baz.quux/key.pem")
                                            (:passphrase . "")))))
```

avodonosov:
> >    At the same time,
> > >        However, even where the ClientHello callback is used, the servername callback is still necessary in order to acknowledge the servername requested by the client.
> Yeah, I take this to mean that you should still use the servername callback for SNI since it is the place where one may return one of those status codes to the client (+SSL-TLSEXT-ERR-OK+, etc.), but that ClientHello should be used for other kinds of extensions.

I understood these doc parts as it is recommended to perform all SNI related actions in the ClientHello callback, but in addition return correct status from the servername callback.

avodonosov:
As for reading the certificate file on every connection, that's what can justify the approach with a separate context for every hostname. In that approach every context could be configured with certificate only once, e.g. by `ssl-ctx-use-certificate-chain-file` and then the ClientHello callback selects corresponding context for the connection.

avodonosov:
For the record. 

I am not committing this solution, because I think it's better to create a separate SSL_CTX for every server name, then in the ClientHello callback lookup the servername coming with SNI extension, and switch to the needed SSL_CTX.

I just didn't have time yet to create an example for that.

I suspect the ClientHello callback will need to use functions like `SSL_client_hello_get0_ext()` and `SSL_client_hello_get1_extensions_present()` to retrieve the incoming server name.

Also,  servername callback will need to be defined, but only to acknowledge the servername to the client (return proper status).

(The documentation quotes and links suggesting this approach are provided in the comments above).

PR #118, opened 2021-03-01
Increase CL-internal buffer size for SSL.

phmarek:
The default value is much too small - it causes too many (unnecessary)
syscalls and so increases latency quite a lot because of additional
threads scheduling.

With the old value of 2K and a real-world quux-hunchentoot easy-handler
that takes about 20msec to create a ~220kB output:

    Running 15s test @ https://localhost:8070/...
      3 threads and 3 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency    71.09ms   22.98ms 183.60ms   80.96%
        Req/Sec    14.30      5.76    30.00     95.19%
      Latency Distribution
         50%   68.34ms
         75%   76.21ms
         90%   91.75ms
         99%  163.99ms
      637 requests in 15.03s, 142.58MB read
    Requests/sec:     42.37
    Transfer/sec:      9.48MB

With `+INITIAL-BUFFER-SIZE+` set higher:

    Running 15s test @ https://localhost:8070/...
      3 threads and 3 connections
      Thread Stats   Avg      Stdev     Max   +/- Stdev
        Latency    39.89ms   21.91ms 151.72ms   75.78%
        Req/Sec    25.35      7.16    60.00     89.06%
      Latency Distribution
         50%   28.86ms
         75%   61.30ms
         90%   67.58ms
         99%   96.53ms
      1140 requests in 15.03s, 255.74MB read
    Requests/sec:     75.85
    Transfer/sec:     17.02MB

Sadly openssl by default uses a 16kB buffer, so there are still
16 `write()` requests on the socket; perhaps the SSL socket buffer size
should be synchronized to `+INITIAL-BUFFER-SIZE+` via `BIO_ctrl`
(`bio-set-fd` in `cl+ssl`) and `BIO_C_SET_BUFF_SIZE` (currently not
available)?

As the content has to be generated in one go (no streaming data!)
any streaming/latency considerations don't apply here; but I guess it
would be much better to change the constant to a special, so that
tuning on a per-socket acceptor base is easily done.

### DIFF
--- a/src/ffi-buffer-all.lisp
+++ b/src/ffi-buffer-all.lisp
@@ -4,7 +4,7 @@
 
 (in-package :cl+ssl)
 
-(defconstant +initial-buffer-size+ 2048)
+(defconstant +initial-buffer-size+ 131072)
 
 (declaim
  (inline


avodonosov:
How do you think, what determines most of the improvement you observe - smaller number of syscalls, or some Lisp<->C ffi interaction?

avodonosov:
For example, if we make +INITIAL-BUFFER-SIZE+ == the size of internal OpenSSL buffer == 16 kB, how does in compare with the 130 kB from the pull request?

phmarek:
Well, the smaller number of syscalls helps to keep the thread on the CPU; if we could get a write size of 256KB, it would get back to close the connection.

With 16kB I get

```
Running 15s test @ https://localhost:8070/...
  3 threads and 3 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
	Latency    90.35ms   25.35ms 175.71ms   70.22%
	Req/Sec    11.47      4.25    20.00     74.58%
  Latency Distribution
	 50%   83.78ms
	 75%  103.71ms
	 90%  128.10ms
	 99%  171.84ms
  497 requests in 15.03s, 111.37MB read
Requests/sec:     33.07
Transfer/sec:      7.41MB
```

so not that much different.


See [here](https://sourceforge.net/p/sbcl/mailman/sbcl-devel/thread/d558233a87790d65c97207cec10ee71b%40marek.priv.at/#msg37225620) for performance data on an unencrypted connection; I hope that (apart from the additional CPU usage, ie. with only _added_ latency but not so much variation!) I can get similar speedups with SSL as well.

avodonosov:
I have just attempted to change the +initial-buffer-size+ to parameter, but found this: 
https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/src/ffi-buffer-clisp.lisp#L42
Thinking how to make it right.

One thing in your explanation that is unclear to me. I understood that beyond 16 kB buffer size the number of system calls will remain constant, because OpenSSL anyways uses on 16 kB buffer internally. If so, why do we see more speedup when growing buffer beyond that?

Maybe that's not system calls, but some Lisp operation, e.g. with-pointer-to-vector data?

phmarek:
> I have just attempted to change the +initial-buffer-size+ to
> parameter, but found this:
> https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/src/ffi-buffer-clisp.lisp#L42
> Thinking how to make it right.

Hmmm...

We could make it

    ( #+cmucl defconstant
      #-cmucl defparameter +initial-buffer-size   131072 )

(but see below)


> One thing in your explanation that is unclear to me. I understood that
> beyond 16 kB buffer size the number of system calls will remain
> constant, because OpenSSL anyways uses on 16 kB buffer internally.

That's what I observed using strace, right.


> If
> so, why do we see more speedup when growing buffer beyond that?
> 
> Maybe that's not system calls, but some Lisp operation, e.g.
> with-pointer-to-vector data?

Well, there are a few factors at play here.

* Basically, the HT handler gets input data, wrangles it in memory,
   and prepares a PDF. So all the CPU use is after receiving and
   before sending.
   PDF size is currently about 240kB.
   I'm testing with a few threads, so CPU contention happens; threads
   get moved on/off CPUs, causing latency variations (bad).

* The optimal case (no SSL): Lisp prepares one big vector (I simply
   reserve 1MB), and writes that big vector to the socket; the kernel
   can just push it out.
   + Only one syscall
   + (Typically) won't get scheduled away in the middle of transmission
   + Lowest possible latency and variation

* The OpenSSL case: Lisp passes a big vector to OpenSSL, which cuts
   it into pieces for encryption, and sends these pieces out one by one.
   ~ I see 16kB blocks being sent in the strace output - better than 4kB
   - gets scheduled away a few times
   - takes quite a bit longer than clear text sending
	(encryption, ie another memory-memory copy)

* The OpenSSL case before my change: same as above, but with 2kB writes
   to the socket
   - so more than 100 syscalls,
   - with much scheduling off the CPU
   - veeeeery slow and high variation in latency



I didn't find the place where 16kB are defined inside openSSL yet;
I tried this:

   (cond
     ...
     ((typep hunchentoot::*hunchentoot-stream* 'cl+ssl::ssl-stream)
      (let* ((ssl (cl+ssl::ssl-stream-handle 
hunchentoot::*hunchentoot-stream*))
             (bio (cl+ssl::ssl-get-wbio ssl))
             (res (cl+ssl::bio-set-fd bio
                     cl+ssl::+bio_c_set_buff_size+
                     cl+ssl::+initial-buffer-size+
                     (cffi:null-pointer))))
	   (print res *trace-output*))
      ;; write via openssl...
	 ))

but that gave me a "0" (failed), this SSL call would have returned
1 for success...


Perhaps you have another idea here?



I guess we want to have the buffer size configurable _somehow_ -
for clisp in the worst case with some flag in *features*,
perhaps :clisp-ssl-big-buffers or :clisp-ssl-small-buffers?

Depending on the number of concurrently active connections
1MB or 16kB might be more appropriate...



Thanks for your time and patience!

avodonosov:
@phmarek , I don't understand what defined the difference between 16 kB buffer and larger buffer - the syscall number should be the same for those situations. Do you have an idea?

Also, why is scheduling the thread away from CPU when it's waiting for OS to send data is bad? Should be good, because the thread is doing nothing anyways, so better to give the CPU to others.

avodonosov:
@phmarek , see also the second commit

phmarek:
> @phmarek , I don't understand what determines the difference between 16 kB buffer and larger buffer - the syscall number should be the same for those situations. Do you have an idea?

Reducing the syscall number is what this about, so we need larger buffers.


> Also, why is scheduling the thread away from CPU when it's waiting for OS to send data is bad? Should be good, because the thread is doing nothing anyways, so better to give the CPU to others.

Well, the point is that individual threads can be freed to work on new requests! I'll try a diagram.


What I'd like to achieve: A thread gets a request, processes it (using some CPU), and then sends the data off to the receiver. The optimal case (with a large enough send buffer and a single syscall) looks like this ("the network" being kernel, hardware interrupts):

```
thread  |work on request 1|work on request 2|work on request 3|
network                   >sends-1          >sends-2          >sends-3
```

Here it doesn't even matter if the same threads gets to work on the next request or if it's another thread - the work is done, only the data needs to be spooled to the client.


With a small send buffer, it looks like that:
```
thread  |work on request 1| write()  write()  write() |
another thread                                        | gets to be active, works on request 2
network                   >1       >2       >3       >4
```

and so the request latency (from a client POV) is much, much higher.


The effect on requests/sec and latency is this (with about 20msec of _pure_ CPU use in the handler, no SSL yet to show the difference between stream buffers of 4KB and 1MB):

```
                |    1 client     |    4 clients
----------------+-----------------+------------------
small buffer    |     19.51req/s  |     21.08req/s 
                |  avg   50.90ms  |  avg   47.38ms
                |  50%   51.53ms  |  50%   60.42ms
                |  75%   68.13ms  |  75%   65.09ms
                |  90%   80.48ms  |  90%   73.14ms
                |  99%   91.87ms  |  99%   84.03ms
----------------|-----------------+------------------
large buffer    |     47.79req/s  |     33.37req/s
                |  avg   20.93ms  |  avg   30.38ms
                |  50%   20.27ms  |  50%   23.96ms
                |  75%   22.16ms  |  75%   38.02ms
                |  90%   23.11ms  |  90%   39.81ms
                |  99%   32.03ms  |  99%   61.99ms
```

- In the single-client use, the reduced number of syscalls more than doubles performance;
- With 4 simultaneous clients, we can increase performance more than half, still.

But more important are the latency numbers -- with a large buffer, the client latencies are reduced, meaning that a smaller number of threads is needed to achieve the same throughput!
Which in turn means second-order effects (less I-cache used, less task-switching, etc.)

With SSL in the mix, having CPU-intensive workload mixed with the syscalls only exaggerates the issue; if the entire result could be encrypted at once and then just pushed out, it would be much better. (No idea whether SSL allows that or requires some application-level acks inbetween??)


As for performance with SSL and current `git HEAD`, with the `*default-buffer-size*` of 2048 (`strace` shows `write()` calls with 2070 bytes each):

```
  6 threads and 6 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    60.28ms   22.06ms 132.02ms   60.36%
    Req/Sec    16.62      6.05    30.00     52.61%
  Latency Distribution
     50%   67.67ms
     75%   77.57ms
     90%   85.81ms
     99%  100.03ms
  991 requests in 10.01s, 223.45MB read
Requests/sec:     98.97
Transfer/sec:     22.31MB
```

With a large buffer (512kB, enough to store the full result; `strace` giving `write` sizes of 16406, ie. about 8 times as much per call):
```
  6 threads and 6 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    49.70ms   21.93ms 115.63ms   57.43%
    Req/Sec    20.07      6.72    40.00     58.93%
  Latency Distribution
     50%   43.08ms
     75%   69.48ms
     90%   84.80ms
     99%   94.93ms
  1203 requests in 10.01s, 270.66MB read
Requests/sec:    120.19
Transfer/sec:     27.04MB
```

As I wrote previously, I'd like to get openssl to do more encryption work at once... but haven't succeeded so far.

Thanks - your patches are a step in the right direction, there's already 20% more throughput with SSL!

avodonosov:
@phmarek , look, with 2 kB buffer you had 
    Requests/sec:     42.37
    Transfer/sec:      9.48MB

with 16 kB buffer you had
    Requests/sec:     33.07
    Transfer/sec:      7.41MB

with much higher buffer (130 kB?) you had
    Requests/sec:     75.85
    Transfer/sec:     17.02MB

The theory about syscals does not explain these numbers. After 16kB the number of syscalls remains constant, so the performance should remain constant if the syscalls are the main factor. But we observe opposite, reduction of syscalls on the change from 2 kB to 16 kB hasn't influenced performance significantly, and then further grows of the buffer increased performance very much, despite systalls do not reduce anymore.

Either syscalls continue reducing when we go beyond 16 kB, or something else determines the performance (passing buffer through ffi is expensive, or some other operations we repeat are expensive).

Also, I doubt syscalls and scheduling can influence the throughput that much, especially when we speak of just 3 simultaneous client threads and 3 connections (be it 3 requests or 9 requests in total simultaneously). I do not specialize in this topic, may be wrong, still, I doubt it.

phmarek:
Well, we agree that less syscalls is better.

Similarly, batching the SSL work (encryption) to encrypting one block right after producing the output, so that we can use _one_ big socket `write()` call (instead of encrypting, writing, encrypting, writing), surely reduces client latency - with a single write being the optimal case here.

Do you have an idea how we could get `(cl+ssl::bio-set-fd bio cl+ssl::+bio_c_set_buff_size+ ...)` to work, so that I can offer numbers for a single syscall as well?

phmarek:
Of course, OpenSSL still uses "only" 16kB writes... perhaps I'll find out how to change that

avodonosov:
@phmarek , 

> Do you have an idea how we could get (cl+ssl::bio-set-fd bio cl+ssl::+bio_c_set_buff_size+ ...) to work, so that I can offer numbers for a single syscall as well?

Assuming you pass parameters correctly, probably the limitation here is the maximum size of a TLS record.

> BIO_set_write_buf_size() sets the write buffer size of BIO b to size. If the size is not initialized a default value is used. This is currently 17K, sufficient for a maximum size TLS record.

https://www.openssl.org/docs/man1.0.2/man3/BIO_ctrl_reset_read_request.html

Of course, we can imagine an implementation where OpenSSL prepares multiple TLS records in one buffer and then writes them all in one system call. But in practice, OpenSSL can probably have an implementation where every TLS record is written  by a separate system call.

One way to study this is to set the OpenSSL buffer to a smaller size than 16 kB. If you then observe smaller arrays passed to the system call, then we conclude the buffer resizing works find, but OpenSSL just does not user buffers larger than one record size.

BTW, if you search online for "SSL buffer size" you will find a number of articles explaining that sometimkes smaller records are better, because TCP/IP traffic gets usually fragmented to packets of around 1400 bytes.  E.g. https://blog.cloudflare.com/optimizing-tls-over-tcp-to-reduce-latency/

If setting smaller OpenSSL buffer will work, you will then be able to differentiate by system calls - use 1300 bytes buffer (or smaller) vs the default 17 kByte buffer will result in more than 10 fold syscall number difference. (As already discussed, your current speedup mostly gained when you go from 16 kB to 512 kB, which has not influence on syscall number).

avodonosov:
@phmarek , regarding the idea that a thread will pick up another request, while in background the data is being sent to network, there are many problems:
- The functions we use are synchronous, they only return after data is sent.
- There are also asynchronous functions, cl+ssl even defines nonblocking-ssl-funcall. But for that to work, there should anyway be a thread that controls the async IO, and OpenSSL does not provide such a thread in background (as I know). That should be an application thread. So there could be a thread pool that builds the PDFs, and after the PDFs is completed, a pair {PDFs, output stream} is added to an output collection that is handled by a single thread, that writes to all sockets using asyn IO.
```
http-request-1 ->
http-request-2 ->   pdf generation pool  -> output thread [pdf-3, pdf-1, pdf-2, ...]
http-request-3 ->
```
- For that to work, the web server should support async request handling. In async mode, after the HTTP request handler returns, the server should not finalize request processing (close the stream or something). Hunchentoot is not designed for this, I believe. Even if you get hold of the hunchentoot output stream, and that stream does not require cleanup after request handler function has returned (that's a direct binary cl+ssl stream, not a flexi-streams and chunga wrapper around it), and keep-alive connection is used, hunchentoot will start counting a timeout after the handler returns, and will close the socket after that timeout, in case no new request was sent from client. If the app haven't sent the complete response by that time, the response will be interrupted in the middle.
- There is no evidence the async mode is worth implementing, till the number of simultaneous connections exceeds the number of threads your process can host (tens of thousands on today's machine -  if we assume 1 MB statically allocated stack, then 10 000 threads will require 10 GB of memory).

Of course, that would be great if you track down the bottleneck and fix it. But practically speaking, that's typical to offload TLS handling from app server to a reverse-proxy or a load balancer in front of it. If you run nginx or haproxy in front of hunchentoot and let them implement TLS, you may have better performance.

avodonosov:
@phmarek , I think if the TCP send buffer is big enough, the application side buffer size does not matter, your thread can be freed anyways.
It will just perform multiple syscalls posting data to the socket; every syscall will place the data to buffer and return immediately.

phmarek:
Hmmm, no, it doesn't seem to work that way.

My "server" has
```
net.ipv4.tcp_rmem = 65536       131072  6291456
net.ipv4.tcp_wmem = 65536       131072  4194304
```

and I simulated a slow client via
```
sudo sysctl net.ipv4.tcp_rmem="4096 4096 4096"
curl --insecure --limit-rate 4096 'https://....'
```

Then an
```
strace  -f -tt -s 20 -e recvfrom,sendto,write,read,futex,close,accept -p <pid>
```

shows
```
08:18:01.273014 accept(9, {sa_family=AF_INET, sin_port=htons(35710), sin_addr=inet_addr("10....")}, [16]) = 12
08:18:01.288762 read(12, "\343\241]\360Bd\v\241\334\207K\177\210n\233)\206s\4c"..., 470) = 470
08:18:01.321450 write(12, "\27\3\3@\21Z`iH\222i\nVD\331aD\36|)"..., 16406) = 16406
... some writes
08:18:36.228210 write(12, "\27\3\3@\21\316\"\221.gL\202(\374]{\16v\312\3"..., 16406) = 942
08:18:36.228496 write(12, "\203\330}\355?\\\345\353\301n\27\\V\242?)\240\227\254\214"..., 15464) = -1 EAGAIN (Die Ressource ist zur Zeit nicht verfügbar)
08:18:40.376915 write(12, "\203\330}\355?\\\345\353\301n\27\\V\242?)\240\227\254\214"..., 15464) = 15464
08:18:40.377527 write(12, "\27\3\3\25\3302\202\316W\224f\347\24Rd8\226\24\222\17"..., 5597) = 3212
08:18:40.378002 write(12, "\365\261\273\315/\207\273kQo2\337\317\20<x(\323\270\f"..., 2385) = -1 EAGAIN (Die Ressource ist zur Zeit nicht verfügbar)
08:18:45.132940 write(12, "\365\261\273\315/\207\273kQo2\337\317\20<x(\323\270\f"..., 2385) = 2385
08:18:45.133485 read(12, 0x7f766000c5a3, 5) = -1 EAGAIN (Die Ressource ist zur Zeit nicht verfügbar)
08:18:56.559629 read(12, "\27\3\3\0\23", 5) = 5
08:18:56.560106 read(12, "\f\326\0\267\350\315)H\362$^\4\251\217\31\3203a\325", 19) = 19
08:18:56.560650 close(12)  = 0
08:18:56.561208 futex(0x100708dd88, FUTEX_WAKE_PRIVATE, 1 <unfinished ...>
```

So a slow connection _will_ slow down the sending thread.
If I use my (default) `net.ipv4.tcp_rmem = 65536        262144   6291456` on the client, the server can push ~140kB before being blocked again.

But by `sysctl 'net.ipv4.tcp_wmem=524288 524288 4194304'` on the server I can get all the writes() to succeed at once -- only waiting the TLS ACK answer still blocks the thread from doing other work, sadly.

So, concluding -- with local `tcp_wmem` settings I can ignore the 16kB SSL buffer size, but the threads still are blocked. (Without SSL, they can start other work immediately.)


Thanks for the change, that one at least reduces the number of FFI calls!

PR #125, opened 2021-05-11
Send shutdown alert to peer when closing SSL stream.

jathd:
Right now, closing an `ssl-stream` doesn't call `SSL_shutdown()`, resulting in a connection that is "non-properly terminated" according to clients such as `gnutls-cli`.


avodonosov:
Thanks, ,merged

PR #126, opened 2021-05-11
Fix return type of `ssl-shutdown`.

jathd:
The foreign function `SSL_shutdown()` returns an `int`, not `void`.


avodonosov:
Thanks, merged

PR #127, opened 2021-05-12
Implement extraction of certificate fingerprint.

jathd:
Implements a new function `certificate-fingerprint` that extracts the fingerprint of a certificate, with accompanying tests.

Issue #61, opened 2018-01-10
Lispworks+OSX high sierra: cl-plus-ssl tries to use boringssl and crashes on "ssl_create_cipher_list"


paulvaneecke:
When using cl-plus-ssl with lispworks on mac osx high sierra (in which Apple replaced openssl with boringssl), the initialize function throws the following error:

;; Causing the error
(asdf:make :cl+ssl)
**(cl+ssl::initialize)**

;; Error
Error: Segmentation violation(11) [code 0] at 7FFF64C5F578
        Foreign code offset #x69 from symbol "ssl_create_cipher_list"
        module "/usr/lib/libboringssl.dylib" [ #x7FFF64BDC000 ]

It seems like it tries to use boringssl anyways.
The error does not occur on the same system with SBCL or CCL; only with lispworks.

Any ideas?

avodonosov:
What native library do SSL and CCL yse? Also boringssl or openssl?

avodonosov:
You can use :CL+SSL-FOREIGN-LIBS-ALREADY-LOADED (https://common-lisp.net/project/cl-plus-ssl/) to load foreing libs yourself - can you use the normal openssl?

paulvaneecke:
On CCL, I get the following:

`(cffi:load-foreign-library "libssl.dylib")
#<FOREIGN-LIBRARY LIBSSL.DYLIB-49974 "libssl.dylib">=

whereas on lispworks

`(cffi:load-foreign-library "libssl.dylib")
#<CFFI:FOREIGN-LIBRARY LIBSSL.DYLIB-48965 "libssl.35.dylib">`

So it appears to be a different actual dylib file, but I don't really understand how these names get resolved...

avodonosov:
Maybe your cl+ssl is too old and doesnt support :cl+ssl-foreign-libs-already-loaded?

What version of quicklisp dist do yoy use?

avodonosov:
That's how native libs are loaded in absenxe of :cl+ssl-foreign-libs-already-loaded: https://github.com/cl-plus-ssl/cl-plus-ssl/blob/08a7524dd9f8ec6c5e3541c2ab62ac82cdfd27a6/src/reload.lisp#L35

avodonosov:
Regarding your (cffi:load-foreign-library "libssl.dylib") above: use absolute path to control exactly what is liaded.

stablecross:
CL-USER 1 > (ql:quickload :cffi)
[...]

PR #128, opened 2021-05-15
Fix tests on ABCL

jathd:
This fixes tests on ABCL, including the "expected failures" that shouldn't have been expected at all.

The problem is this: when testing, we preload "libcrypto" and "libssl", and prevent the library named by the symbol `libssl` from being loaded, but we still specify `:library libssl` in foreign function definitions.

This problem manifests on ABCL because its CFFI implementation actually uses the `:library` argument, whereas other implementations seem to simply ignore it. For instance, in cffi-sbcl you can find:
```
(defmacro %foreign-funcall (name args &key library convention)
  "Perform a foreign function call, document it more later."
  (declare (ignore library convention))
...)
```

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -111,7 +111,8 @@ variants if you have use cases for them.)"
                *cl+ssl-ssl-foreign-function-names*
                :test 'equal)
      (defcfun-versioned (:since ,since :vanished ,vanished)
-         ,(append name-and-options '(:library libssl))
+         ,(append name-and-options
+                  #-cl+ssl-foreign-libs-already-loaded '(:library libssl))
        ,@body)))
 
 (defmacro define-ssl-function (name-and-options &body body)
@@ -144,7 +145,7 @@ variants if you have use cases for them.)"
          ;; mechanism with possibility for user to specify value
          ;; for the :library option.
          ,(append name-and-options
-                  #+(and (or abcl lispworks) darwin) '(:library libcrypto))
+                  #-cl+ssl-foreign-libs-already-loaded '(:library libcrypto))
        ,@body)))
 
 (defmacro define-crypto-function (name-and-options &body body)
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -125,25 +125,6 @@ variants if you have use cases for them.)"
               *cl+ssl-crypto-foreign-function-names*
               :test 'equal)
      (defcfun-versioned (:since ,since :vanished ,vanished)
-         ;; On Darwin, LispWorks has boringssl always loaded
-         ;; (https://github.com/cl-plus-ssl/cl-plus-ssl/issues/61),
-         ;; ABCL somehow has libressl loaded
-         ;; (https://github.com/cl-plus-ssl/cl-plus-ssl/pull/89),
-         ;; and when we load openssl and declare ffi functions
-         ;; without explicitly specifying the :library option,
-         ;; some foreign symbols are resolved as boringssl / libressl symbols,
-         ;; others are resolved as openssl functions.
-         ;; This mix results in failures, of course.
-         ;; We fix these two implementations by passing the :library option.
-         ;; Not for other implementations because this may be
-         ;; incompatible with :cl+ssl-foreign-libs-already-loaded
-         ;; but these two implementations just break without
-         ;; that, so it's better to possibly sacrify the
-         ;; :cl+ssl-foreign-libs-already-loaded (we haven't tested)
-         ;; than have them broken completely.
-         ;; TODO: extend the :cl+ssl-foreign-libs-already-loaded
-         ;; mechanism with possibility for user to specify value
-         ;; for the :library option.
          ,(append name-and-options
                   #-cl+ssl-foreign-libs-already-loaded '(:library libcrypto))
        ,@body)))
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,11 +135,10 @@ script:
                 (format t \"abcl-asdf::*default-repository* assigned the HTTPS URL.~%\"))'
          -e '(ql:quickload :cffi)'
          -e '(format t \"cffi loaded.~%\")'
-         -e '(cffi:load-foreign-library \"test/run-on-many-lisps-and-openssls/openssl-releases/bin/'$OPENSSL-${BITS}bit'/lib/libcrypto.so\")'
-         -e '(format t \"libcrypto.so loaded.~%\")'
-         -e '(cffi:load-foreign-library \"test/run-on-many-lisps-and-openssls/openssl-releases/bin/'$OPENSSL-${BITS}bit'/lib/libssl.so\")'
-         -e '(format t \"libssl.so loaded.~%\")'
-         -e '(pushnew :cl+ssl-foreign-libs-already-loaded *features*)'
+         -e '(ql:quickload :cl+ssl/config)'
+         -e '(format t \"cl+ssl/config loaded.~%\")'
+         -e '(cl+ssl:define-libcrypto-path \"test/run-on-many-lisps-and-openssls/openssl-releases/bin/'$OPENSSL-${BITS}bit'/lib/libcrypto.so\")'
+         -e '(cl+ssl:define-libssl-path \"test/run-on-many-lisps-and-openssls/openssl-releases/bin/'$OPENSSL-${BITS}bit'/lib/libssl.so\")'
          -e '(ql:quickload :cl+ssl) ;; load cl+ssl separately from cl+ssl.test only because cl+ssl.test can not be loaded in the :invert readtable-case due to its dependency ironclad, as of 2019-10-20'
          -e '(format t \"cl+ssl loaded.~%\")'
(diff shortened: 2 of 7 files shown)


jathd:
I just noticed that there was already a mention of this in a comment in `define-crypto-function-ex`:
```
         ;; On Darwin, LispWorks has boringssl always loaded
         ;; (https://github.com/cl-plus-ssl/cl-plus-ssl/issues/61),
         ;; ABCL somehow has libressl loaded
         ;; (https://github.com/cl-plus-ssl/cl-plus-ssl/pull/89),
         ;; and when we load openssl and declare ffi functions
         ;; without explicitly specifying the :library option,
         ;; some foreign symbols are resolved as boringssl / libressl symbols,
         ;; others are resolved as openssl functions.
         ;; This mix results in failures, of course.
         ;; We fix these two implementations by passing the :library option.
         ;; Not for other implementations because this may be
         ;; incompatible with :cl+ssl-foreign-libs-already-loaded
         ;; but these two implementations just break without
         ;; that, so it's better to possibly sacrify the
         ;; :cl+ssl-foreign-libs-already-loaded (we haven't tested)
         ;; than have them broken completely.
         ;; TODO: extend the :cl+ssl-foreign-libs-already-loaded
         ;; mechanism with possibility for user to specify value
         ;; for the :library option.
```
Which means that this PR is a band aid (still useful though, because this problem is what's making travis fail in PR #127, for instance), but it would be nice to have a more robust mechanism for choosing which library is loaded when testing.

avodonosov:
How do you know that is the cause?15.05.2021, 14:03, "jathd" ***@***.***>:
I just noticed that there was already a mention of this in a comment in define-crypto-function-ex:
         ;; On Darwin, LispWorks has boringssl always loaded
         ;; (https://github.com/cl-plus-ssl/cl-plus-ssl/issues/61),
         ;; ABCL somehow has libressl loaded
         ;; (https://github.com/cl-plus-ssl/cl-plus-ssl/pull/89),
         ;; and when we load openssl and declare ffi functions
         ;; without explicitly specifying the :library option,
         ;; some foreign symbols are resolved as boringssl / libressl symbols,
         ;; others are resolved as openssl functions.
         ;; This mix results in failures, of course.
         ;; We fix these two implementations by passing the :library option.
         ;; Not for other implementations because this may be
         ;; incompatible with :cl+ssl-foreign-libs-already-loaded
         ;; but these two implementations just break without
         ;; that, so it's better to possibly sacrify the
         ;; :cl+ssl-foreign-libs-already-loaded (we haven't tested)
         ;; than have them broken completely.
         ;; TODO: extend the :cl+ssl-foreign-libs-already-loaded
         ;; mechanism with possibility for user to specify value
         ;; for the :library option.

Which means that this PR is a band aid (still useful though, because this problem is what's making travis fail in PR #127, for instance), but it would be nice to have a more robust mechanism for choosing which library is loaded when testing.

src/ffi.lisp
```diff
@@ -111,7 +111,10 @@ variants if you have use cases for them.)"
                *cl+ssl-ssl-foreign-function-names*
                :test 'equal)
      (defcfun-versioned (:since ,since :vanished ,vanished)
-         ,(append name-and-options '(:library libssl))
+         ,(append name-and-options
+                  ;; If foreign libraries were already loaded, the library
+                  ;; cl+ssl::libssl is never loaded, so don't use it.
+                  #-:cl+ssl-foreign-libs-already-loaded '(:library libssl))
```
avodonosov, line 117: We should use run-time check instead of the read time check.

jathd:
Okay, so here's a proposal that removes all the reader conditionals in the source code.

The idea is to always define libraries `libcrypto` and `libssl`, even when testing, so we can always use the options `:library libcrypto` and `:library libssl`.

avodonosov:
By run-time check I simply meant `(member :cl+ssl-foreign-libs-already-loaded *features*)`, but the new proposal is a better direction. Only I will need to think through the backward compatibility.

jathd:
> By run-time check I simply meant (member :cl+ssl-foreign-libs-already-loaded *features*)

Yes, I also thought about using a special variable that we could bind around the loading of the whole system, but I didn't see a convenient way to do that using ASDF.
This proposal seemed like a decent compromise.

What are your worries concerning backward compatibility?

avodonosov:
> What are your worries concerning backward compatibility?

The people who use `:cl+ssl-foreign-libs-already-loaded` do not expect the `(use-foreign-library ...)` to happen, for example.

jathd:
Ah, that's true. I hadn't noticed that this feature was documented.

jathd:
About the dynamic check `(member :cl+ssl-foreign-libs-already-loaded *features*)`: I don't think there's any guarantee that this is more dynamic, because it occurs inside a macro-expander. The compiler could expand all macros at compile-time, and then loading the compiled file won't obey this check. (I would actually expect any reasonable compiler to do exactly this.)

The problem is that with `cffi`, the `:library` keyword argument is not evaluated, and neither are the paths in `define-foreign-library`, so I don't think there is a simple way to decide at runtime what library to use.

jathd:
Here's a way to dynamically define the path of libss and libcrypto.
For instance:
```
(ql:quickload :cl+ssl/config)
(cl+ssl:define-libssl-path "libssl.so.1.0.0")
(ql:quickload :cl+ssl)
```
The variables `*libssl-override*` and `*libcrypto-override*` are set to true when the `define-*-path` macros are used, and so the default definitions of the libraries are not loaded.

On some implementations, you can redefine the paths and reload cl+ssl to use the new library. For instance on ABCL:
```
(ql:quickload :cl+ssl)
(cl+ssl::compat-openssl-version)
```
prints `269488319` on my machine (corresponding to version 1.1). If I then do
```
(cl+ssl:define-libcrypto-path "libcrypto.so.1.0.0")
(cl+ssl:reload)
(cl+ssl::compat-openssl-version)
```
I get `268443999` (corresponding to version 1.0.0).

On other implementations (SBCL, for instance) this doesn't work and you have to restart the image, because the ffi implementation caches the results of foreign functions lookup, so whatever library you load first is always used subsequently.

avodonosov:
If you want. we can merge #127 by adding the new failed test to the list of expected ABCL failures. So that PR does not wait for the current PR - I need more time to review it.

jathd:
Sure, why not. I'll go do that.

jathd:
(By the way, if this looks good to you, I'll add some documentation and rewrite history into something sensible.)

jathd:
Brain fart: that last commit abound backward compatibility is not actually backward compatible with the previous behaviour.

OK, this is actually a bit tricky. What is the expected behaviour that should happen when feature `:cl+ssl-foreign-libs-already-loaded` is enabled? (Apart from not calling `use-foreign-library` on libcrypto nor libssl. )
Should we keep the old behaviour of not using the `:library` keyword, except on abcl and lispworks on darwin? That's probably the solution that is maximally backwards compatible.

jathd:
(Force pushed to make a cleaner branch and rebase on top of current master.)

Here's the situation with this latest proposal:

- If feature `cl+ssl-foreign-libs-already-loaded` is present, we don't load `libssl` and `libcrypto` (that was already the case) and we don't use the options `:library libssl` and `:library libcrypto` in foreign function definitions (this is new in commit dd79dc0). 
This change is obviously backwards compatible on implementations that don't care about the `:library` option (SBCL and CCL, maybe others).
On implementations that do care about the `:library` option (ABCL, probably Lispworks), the feature was broken anyway (as evidenced by the "expected failures" for tests on ABCL) because the options `:library libssl` and `:library libcrypto` were used even though `libssl` and `libcrypto` were not loaded.

- If feature `cl+ssl-foreign-libs-already-loaded` is not present, the options `:library libssl` and `:library libcrypto` are always used, on all implementations. This was already the case for `:library libssl`, but `:library libcrypto` was only used with ABCL and Lispworks on Darwin (to prevent a mix-up with other libraries that were already preloaded by these implementations).
I don't see how any backward compatibility issue could arise. According to a comment (the one removed in commit a03d92b), the only reason this option wasn't used all the time was to allow people to override libraries using `cl+ssl-foreign-libs-already-loaded`. This is still possible.

- Instead of `cl+ssl-foreign-libs-already-loaded`, a better way to override the libraries is provided through the macros `define-libssl-path` and `define-libcrypto-path`. When one of the is used, it defines the library `libssl` or `libcrypto` and prevents the default definition from being loaded.
All this is done dynamically: no more reader conditionals, no more recompilation.

avodonosov:
Merged

avodonosov:
@jathd , sorry for the delay, I wanted to run Travis Ci tests in both modes: `:cl+ssl-foreign-libs-already-loaded` and `cl+ssl/config`.
Implemented that finally.

Also, I did a little change - gave the `cl+ssl/config` its own separate package. 
So instead of 
```common-lisp
(cl+ssl:define-libssl-path ...)
```
we use
```common-lisp
(cl+ssl/config:define-libssl-path ...)
```

See https://github.com/cl-plus-ssl/cl-plus-ssl/commit/bfe3d72c1a1511dacf1a066e2195d6b7dbb868a8

Thank you for the great contribution!

PR #130, opened 2021-06-02
Add INPUT-WAIT function implementation for Lispworks

svetlyak40wt:
This commit fixes these nasty warnings:

    Warning: cl+ssl::input-wait is not implemented for this lisp, but a non-blocking stream is encountered

I wasn't able to imagine a code example to play with output and to implement an OUTPUT-WAIT counterpart.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -921,7 +921,18 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
                                  :timeout timeout
                                  :whostate "cl+ssl waiting for input")))
 
-#-(or clozure-common-lisp sbcl allegro)
+#+lispworks
+(defun input-wait (stream fd deadline)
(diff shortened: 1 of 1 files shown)


avodonosov:
Thank you, merged. Maybe you can also implement output-wait?

svetlyak40wt:
@avodonosov can you help me with a code snippet that will use `cl+ssl::output-wait`? Having a code to experiment with, I'll be able to implement this function for Lispworks.

avodonosov:
@svetlyak40wt , I don't have a tested code, but I think output-wait will be called if one sends a relatively large data, and network is slow, and / or, the TCP buffer is smaller than the data sent.

Some commands how to simulate slow network and set TCP buffer size can be found here: https://github.com/cl-plus-ssl/cl-plus-ssl/pull/118#issuecomment-797295172

PR #133, opened 2021-07-25
Bio for recent openssl

zellerin:
I would like to get some feedback on idea of fixing bio.
My - most likely incomplete - understanding that I got from https://github.com/cl-plus-ssl/cl-plus-ssl/issues/72 is that:
- bio is broken due to ssl moving forward
- but it is not used anyway for anything on significant platforms so noone really cares
- it would be nice to have it fixed in  backward compatible way

This set of patches fixes bio to the point that it works in basic situations as show in the attached test file (test/bio.lisp), and can read/write pem file properly (actually, this was my starting point).

My questions are:
- is bio supposed to be used in some way elsewhere, and how can i be tested whether it works?
- is use of bio for reading pem etc in scope of cl+ssl, or should it really be a separate package?
- do you still think support of old versions of ssl is needed (this is probably easy with two disctint bio methods, but I would argue for dropping old ssl support)
- is there some proper way to run tests, and should I bother with them? Apparently, the (run-all-tests) expect some pem files in /home/david that my machine does not have?
- what else have I missed and any other guidance (esp. in area of create and destroy function - why do we have it - and BIO_FLAGS, or stylistic)?
 
There are still some rough edges in the code, but I wanted to get some feedback before cleaning up more.


avodonosov:
Hi, thank you for the effort.

>    is bio supposed to be used in some way elsewhere, and how can i be tested whether it works?

For network streams, that are used for TLS communications. See the `unwrap-stream-p` parameter of the public functions.

The `unwrap-stream-p` means "give OpenSSL the file descriptor of the stream, instead of a Lisp BIO".
When the parameter is true, which is the default, AND the `stream-fd` is implemented for the current Lisp Implementation, the lisp Lisp BIO is not used.

Up until #98 (the commit cc608f15a1dfa8fe7bcec4bd359f6f56b9e59709) the Lisp BIOs were used by ABCL.

>    is use of bio for reading pem etc in scope of cl+ssl, or should it really be a separate package?
Why would be that interesting to the users, what are the use cases?

>    do you still think support of old versions of ssl is needed (this is probably easy with two disctint bio methods, but I would argue for dropping old ssl support)

Yes, please, keep 100% backward compatibility, unless there are good reasons. Same as you, I hope that will be easy - just introduce parallel code instead of modifying existing functions.

>    is there some proper way to run tests, and should I bother with them? Apparently, the (run-all-tests) expect some pem files in /home/david that my machine does not have?

There are Travis CI tests: https://github.com/cl-plus-ssl/cl-plus-ssl/blob/master/.travis.yml

>    what else have I missed and any other guidance (esp. in area of create and destroy function - why do we have it - and BIO_FLAGS, or stylistic)?

I will try to comment more later.

zellerin:
Hi and thanks for reaction.

I tried to set `unwrap-stream-p` to nil (see last `*TEST ONLY*` commit) and run on my machine with 5am; the tests passed. I was unable to run travis CI even for the original repository due to reasons not related to my changes (sbcl-bin/2.1.7 failure due to missing GLIBC_2.28 for all values of LISP, not sure what I do wrong). I am not familiar with Travis, so I will postpone this.

Would it be OK to have a `*default-unwrap-stream*` variable to control the default?

I will come back to the other points later.

Best regards,
Tomas

avodonosov:
> Would it be OK to have a `*default-unwrap-stream*` variable to control the default?

Yes, absolutely.

avodonosov:
Travis CI is fixed now. I am looking also into migration to github actions.

zellerin:
Hi,

the code now builds and test run on sbcl and both old and new lisp. Please note that the Travis run above is not relevant where it passes (see below)

Summary of changes:
- New variable - `*default-unwrap-stream-p*` is used, defaulting to t
- New variable - `*bio-methods-have-opaque-slots*` is used to determine how to access bio, default is t for 1.1.0 and above, nil for below.
- Even new bio code has some changes (two new callbacks) and passes tests in test/bio

Known open points for bio itself:
- Travis fails. Most likely using feature instead of variable for *bio-methods-have-opaque-slots* would be more appropriate to prevent compile time errors where functions or slots are not defined and cffi:defcstruct not being top level, so I would go this way.
- Travis config is not updated, so it runs with T for *default-unwrap-stream-p*. Options I can see are to this as another variable in Travis matrix, to change the script to run all tests with both options (with the variable bound to t and then to nil), or duplicating in all relevant tests where variable is used. I could use help with Travis, thought. 
- I am not sure where to make a cut on libressl version (and I cant find that easily).

avodonosov:
@zellerin , what was failing before you started to use feature?

src/bio.lisp
```diff
@@ -56,13 +59,25 @@
...
       (setf (slot 'ctrl) (cffi:callback lisp-ctrl))
-      (setf (slot 'create) (cffi:callback lisp-create))
+      (setf (slot 'create) (cffi:callback lisp-create-slots))
       (setf (slot 'destroy) (cffi:callback lisp-destroy))
       (setf (slot 'callback-ctrl) (cffi:null-pointer)))
     m))
 
+#+bio-opaque-slots
+(defun make-bio-lisp-method ()
+  (let ((m (bio-meth-new (load-time-value (bio-new-index)) "lisp")))
```
avodonosov, line 71: where is the `bio-meth-new` defined?

avodonosov: nevermind, I see it

zellerin:
> @zellerin , what was failing before you started to use feature?
Can you access https://app.travis-ci.com/github/cl-plus-ssl/cl-plus-ssl/builds/234769530?
- `Error: Unknown CFFI type (:STRUCT BIO-METHOD)` for ccl and ecl, both new and old openssl (probably because struct definition not toplevel)
- `(Attempt to call  #("BIO_get_new_index" 4155758080 0 2 4155758080 0) for which the definition has not yet been (or is no longer) loaded.)` for allegro and old Openssl
- `CFFI::UNDEFINED-FOREIGN-TYPE-ERROR` for abcl

> It's better to not conditionalize the cffi structure definitions. Can we keep the structures defined always? As long as the old code is not called, the we don't try to access the structure fields, so there should be no problem.

Aye. It should work, and maybe fix the errors better.

avodonosov:
I am experimenting with this branch to address https://github.com/edicl/hunchentoot/issues/189 where hunchentoot threads are leaked endlessly waiting for requests over cl+ssl stream,  because when cl+ssl passes stream file descriptor to OpenSSL, the lisp stream read / write timeouts are not obeyed.

With LispBIO, the OpenSSL delegates IO to original lisp stream, where timeouts are in effect, and that helps, at least on CCL.

But we will need to capture all `serious-conditions`  inside all the the callbacks. Because in the current form, the timeout error signalled by the stream to the `lisp-read` initiates a non-local control transfer (https://github.com/edicl/hunchentoot/blob/master/acceptor.lisp#L367, https://github.com/edicl/hunchentoot/blob/master/util.lisp#L377) across C call, which is not good, as explained in CFFI tutorial: https://common-lisp.net/project/cffi/manual/html_node/Tutorial_002dCallbacks.html

`process-connection :around` -> `SSL_accept` -> `lisp-read` -> `return to process-connection :around'  

Example backtrace:

```
[2021-08-10 06:37:24 [ERROR]] Error while processing connection: Error #<USOCKET:TIMEOUT-ERROR #x302001FDE3AD>
 (7FEE39162430) : 0 (PRINT-CALL-HISTORY :CONTEXT NIL :PROCESS NIL :ORIGIN NIL :DETAILED-P NIL :COUNT 1152921504606846975 :START-FRAME-NUMBER 0 :STREAM #<STRING-OUTPUT-STREAM  #x302001FDE37D> :PRINT-LEVEL 2 :PRINT-LENGTH 5 :PRINT-STRING-LENGTH :DEFAULT :SHOW-INTERNAL-FRAMES NIL :FORMAT :TRADITIONAL) 965
 (7FEE391625C8) : 1 (PRINT-BACKTRACE-TO-STREAM #<STRING-OUTPUT-STREAM  #x302001FDE37D>) 85
 (7FEE391625F8) : 2 (GET-BACKTRACE) 397
 (7FEE39162660) : 3 (FUNCALL #'#<(:INTERNAL (HUNCHENTOOT:PROCESS-CONNECTION :AROUND (HUNCHENTOOT:ACCEPTOR T)))> #<USOCKET:TIMEOUT-ERROR #x302001FDE3AD>) 109
 (7FEE391626A0) : 4 (SIGNAL #<USOCKET:TIMEOUT-ERROR #x302001FDE3AD>) 1037
 (7FEE391626F0) : 5 (%ERROR #<USOCKET:TIMEOUT-ERROR #x302001FDE3AD> (:SOCKET NIL) 17582642087139) 117
 (7FEE39162718) : 6 (HANDLE-CONDITION #<INPUT-TIMEOUT #x302001FDE44D> NIL NIL) 253
 (7FEE39162740) : 7 (SIGNAL #<INPUT-TIMEOUT #x302001FDE44D>) 1037
 (7FEE39162790) : 8 (%ERROR #<INPUT-TIMEOUT #x302001FDE44D> (:STREAM #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD>) 17582642087159) 117
 (7FEE391627B8) : 9 (FD-STREAM-ADVANCE #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD> #S(CCL::IOBLOCK :STREAM #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD> :UNTYI-CHAR NIL ...) T) 869
 (7FEE39162830) : 10 (%IOBLOCK-READ-U8-BYTE #S(CCL::IOBLOCK :STREAM #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD> :UNTYI-CHAR NIL ...)) 229
 (7FEE39162868) : 11 (READ-BYTE #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD> T NIL) 285
 (7FEE391628B0) : 12 (FUNCALL #'#<CFFI-CALLBACKS::|CL+SSL::LISP-READ|> 17582641654000) 933
 (7FEE39162910) : 13 (%PASCAL-FUNCTIONS% 8 17582641654000) 405
 (7FEE391629D0) : 14 (SSL-ACCEPT #<A Foreign Pointer #x7FEE2005A4B0>) 125
 (7FEE391629E8) : 15 (MAKE-SSL-SERVER-STREAM #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD> :CERTIFICATE "/home/anton/my/prj/hunchentoot/my-work-cert.pem" :KEY "/home/anton/my/prj/hunchentoot/my-work-key.pem" :PASSWORD NIL :METHOD NIL :EXTERNAL-FORMAT NIL :CLOSE-CALLBACK NIL :UNWRAP-STREAM-P NIL :CIPHER-LIST "ALL" :BUFFER-SIZE 2048 :INPUT-BUFFER-SIZE 2048 :OUTPUT-BUFFER-SIZE 2048) 1605
 (7FEE39162B10) : 16 (FUNCALL #'#<#<STANDARD-METHOD HUNCHENTOOT:INITIALIZE-CONNECTION-STREAM (HUNCHENTOOT:SSL-ACCEPTOR T)>> #<SSL-ACCEPTOR (host *, port 1448)> #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/6) #x302001FAC6CD>) 173
 (7FEE39162B38) : 17 (%%CNM-WITH-ARGS-COMBINED-METHOD-DCODE #(#<STANDARD-GENERIC-FUNCTION HUNCHENTOOT:INITIALIZE-CONNECTION-STREAM #x302001D677DF> (#<#> #<#>) NIL NIL (#<#> #<#>)) 17582642087284 NIL) 1205
 (7FEE39162BB8) : 18 (FUNCALL #'#<#<STANDARD-METHOD HUNCHENTOOT:PROCESS-CONNECTION (HUNCHENTOOT:ACCEPTOR T)>> #<SSL-ACCEPTOR (host *, port 1448)> #<USOCKET:STREAM-USOCKET #x302001FAC3DD>) 517
 (7FEE39162C58) : 19 (%CALL-NEXT-METHOD (NIL #<STANDARD-METHOD HUNCHENTOOT:PROCESS-CONNECTION #> . 17582642087353)) 1069
 (7FEE39162CD0) : 20 (FUNCALL #'#<#<STANDARD-METHOD HUNCHENTOOT:PROCESS-CONNECTION :AROUND (HUNCHENTOOT:ACCEPTOR T)>> #<SSL-ACCEPTOR (host *, port 1448)> #<USOCKET:STREAM-USOCKET #x302001FAC3DD>) 701
 (7FEE39162D68) : 21 (%%STANDARD-COMBINED-METHOD-DCODE (#<STANDARD-METHOD HUNCHENTOOT:PROCESS-CONNECTION :AROUND #> #<STANDARD-METHOD HUNCHENTOOT:PROCESS-CONNECTION #>) 17582642087353) 773
 (7FEE39162DE0) : 22 (FUNCALL #'#<(:INTERNAL HUNCHENTOOT::PROCESS-CONNECTION% (HUNCHENTOOT::HANDLE-INCOMING-CONNECTION% (HUNCHENTOOT:ONE-THREAD-PER-CONNECTION-TASKMASTER T)))> #<SSL-ACCEPTOR (host *, port 1448)> #<USOCKET:STREAM-USOCKET #x302001FAC3DD>) 181
 (7FEE39162E40) : 23 (FUNCALL #'#<(:INTERNAL BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS)>) 845
 (7FEE39162E68) : 24 (RUN-PROCESS-INITIAL-FORM #<PROCESS hunchentoot-worker-127.0.0.1:38968(72) [Active] #x302001FABE0D> (#<COMPILED-LEXICAL-CLOSURE # #x302001FABB9F>)) 717
 (7FEE39162EF0) : 25 (FUNCALL #'#<(:INTERNAL (CCL::%PROCESS-PRESET-INTERNAL (PROCESS)))> #<PROCESS hunchentoot-worker-127.0.0.1:38968(72) [Active] #x302001FABE0D> (#<COMPILED-LEXICAL-CLOSURE # #x302001FABB9F>)) 581
 (7FEE39162F98) : 26 (FUNCALL #'#<(:INTERNAL CCL::THREAD-MAKE-STARTUP-FUNCTION)>) 277
```

zellerin:
Hi Anton,

I can try to catch serious conditions (see above). However, the question is how to communicate to the caller what happened, and what specific conditions to catch (and iirc, timeout condition specifically is named differently in each lisp implementation).

Not sure what the right thing is (use global vars or retry-code slot of the bio), and not sure where this should be read. Some discussion in the commit.

Do you have simple test case for the timeout problem so that I can test?

zellerin:
The remaining Travis errors are mostly (all?) in my new tests that may be wrong and cover previously new callback as well. I will fix them eventually (or remove), but the old functionality should work now across platforms and openssl versions.

zellerin:
So everything but abcl passes now. The abcl error is not exactly helpful and I have minimal experience with that one :(

src/bio.lisp
```diff
@@ -203,9 +203,8 @@
             do
                (setf char (read-byte *socket*)
                      exit (= char 10))
-               (unless exit
-                 (setf (cffi:mem-ref buf :unsigned-char i) char)
-                 (incf i))))
+               (setf (cffi:mem-ref buf :unsigned-char i) char)
+               (incf i)))
```
avodonosov, line 207: You decided to count the newline?

The docs may be interpretted that way, but 
OpenSSL implementanion uses `strlen` on the buffer, thus not counting newline.
https://github.com/openssl/openssl/blob/master/crypto/bio/bss_file.c#L347

zellerin: I did, based on:
https://www.openssl.org/docs/man1.1.1/man3/BIO_gets.html

The returned string is always NUL-terminated and the '\n' is preserved if present in the input data.
...
The trailing NUL is not included in the length returned by BIO_gets().
...and section HISTORY:
BIO_gets() on 1.1.0 and older when called on BIO_fd() based BIO does not keep the '\n' at the end of the line in the buffer._

avodonosov: Right, So the semantics is based on `fgets`, which, unlike `gets`, inclues the new newline character into the `buf`.

src/bio.lisp
```diff
@@ -104,14 +173,52 @@
...
+                       (or (null *partial-read-p*) (listen *socket*)))
+            do
+               (setf char (read-byte *socket*)
+                     exit (= char 10))
+               (setf (cffi:mem-ref buf :unsigned-char i) char)
+               (incf i)))
+      (serious-condition ()
+        (clear-retry-flags bio)))
+    (unless (>= i n)
+      (setf (cffi:mem-ref buf :unsigned-char i) 0))
```
avodonosov, line 211: We should always 0-terminate. This also means loop while i <  n - 1.

src/x509.lisp
```diff
@@ -297,3 +297,16 @@ designator for the digest algorithm to use (it defaults to SHA-1)."
                  :message "failed to compute fingerprint of certificate"
                  :queue (read-ssl-error-queue))))
       fingerprint)))
+
+(defun x509-cert-from-pem (pem)
+  (with-bio-input-from-string (bio pem) (pem-read-x509 bio 0 0 0))
+  pem)
```
avodonosov, line 303: Returning the original input string? Most likely last line is not needed.

zellerin: Right. I have it fixed locally, but I keep additional changes in x509 until core is fixed. I consider I would remove this from this patch/final pull request and left only the core bio fix - that is itself  larger than I originally thought. But your call.

src/bio.lisp
zellerin: Agreed.

zellerin:
Thanks for your effort in review and merge. I can see there was quite a few more things you had to change.

avodonosov:
@zellerin , thank you very much for the valuable contribution. 

Also, during review I learned finally how the Lisp BIOs work.

I could merge your code right away, and improve on top of that in the future.
Because BIOs are disabled by default. And when enabling them hunchentoot worked ok at least in my simple tests.

Some of my changes are cosmetic or pure theoretical, some are substantial, like reporting Lisb BIO error details. I found answers to all the questions we had in the above discussion, I think, if you will have time please review.

Don't worry about ABCL failures, they will be fixed by #128.

You asked how to reproduce the timeout.
```common-lisp

;; make your version of cl+ssl available to quicklisp:
(pushnew "~/prj/cl+ssl/cl-plus-ssl/" asdf:*central-registry* :test 'equal)

(ql:quickload :hunchentoot)

(setq cl+ssl::*default-unwrap-stream-p* nil)

;; When running first time, generate private key files like this:
;;
;; ```sh
;;
;;     cd ~/prj/hunchentoot/
;;     openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout my-work-key.pem -out my-work-cert.pem
;;
;; ```
(defparameter *srv* (make-instance 'hunchentoot:ssl-acceptor
                                   :ssl-privatekey-file "~/prj/hunchentoot/my-work-key.pem"
                                   :ssl-certificate-file "~/prj/hunchentoot/my-work-cert.pem" 
                                   :port 1443
                                   :read-timeout 10
                                   :write-timeout 10))

(hunchentoot:start *srv*)

;; Browse at https://localhost:1443 to see if it's working.


;; To simulate a timeout:
;;
;; ```sh
;;
;;     telnet localhost 1443
;;
;; ```
;;
;; Use slime-list-threads to observe a hunchentoot-worker thread waiting for
;; request on the new connection.
;;
;; Observe hunchentoot logs in to standard output to notice how timeout error is detected.
;; After that the thread hunchentoot-worker thread should disappear, the telnet disconnected.



(hunchentoot:stop *srv*)
``` 
In the current version, when timeout occurs, hunchentoot logs are relatively understandable error:

```
[2021-08-14 21:29:43 [ERROR]] Error while processing connection: A failure in the SSL library occurred on handle #<A Foreign Pointer #x7FB0BC069C70> (return code: 1).
140397467358976:error:01000044:unknown library:unknown library:internal error:cl+ssl/src/bio.lisp:0:Unexpected SERIOUS-CONDITION in the Lisp BIO: Input timeout on #<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/10) #x302004B1FC2D>
```

zellerin:
I did have a look, and it was interesting to see the other changes, especially error handling.

I tried some of my client-side code, and found out that 
- I get errors where there were none previously. Not bug in the bio code, but bio requires underlying stream to be binary (this is logical, but without bio a character stream is silently accepted). Should we check when underlying stream is first seen and warn?
- Following was helpful in finding the error. Not sure if it is fully save across cffi, though.

```
modified   src/bio.lisp
@@ -173,6 +173,8 @@
 ;;; and we only consider unexpected errors here.
 
 (defparameter *file-name* (cffi:foreign-string-alloc "cl+ssl/src/bio.lisp"))
+(defvar *last-callback-error* nil
+  "Error from callback. Used and consumed by error handling code.")
 
 (defun put-to-openssl-error-queue (condition)
   (ignore-errors
@@ -180,8 +182,8 @@
     ;; TODO: starting from OpenSSL 3.0.0 use ERR_raise_data instead
     ;; of the ERR_put_error / ERR_add_error_data combination.
 
-
-    (err-put-error +err_lib_none+ 0 +err_r_internal_error+ *file-name* 0)
+   (setq *last-callback-error* condition)
+   (err-put-error +err_lib_none+ 0 +err_r_internal_error+ *file-name* 0)
 
     #-cffi-sys::no-foreign-funcall ; because err-add-error-data is a vararg function
     (let ((err-msg (format nil
modified   src/ffi.lisp
@@ -878,9 +878,10 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
 (declaim (inline ensure-ssl-funcall))
 (defun ensure-ssl-funcall (stream handle func &rest args)
   (loop
-     (let ((nbytes
-            (let ((*socket* (ssl-stream-socket stream))) ;for Lisp-BIO callbacks
-              (apply func args))))
+    (let* ((*last-callback-error* nil)
+           (nbytes
+             (let ((*socket* (ssl-stream-socket stream))) ;for Lisp-BIO callbacks
+               (apply func args))))
        (when (plusp nbytes)
          (return nbytes))
        (let ((error (ssl-get-error handle nbytes)))
@@ -894,7 +895,9 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
                          (ssl-get-fd handle)
                          (ssl-stream-deadline stream)))
            (t
-            (ssl-signal-error handle func error nbytes)))))))
+            (if *last-callback-error*
+                (error *last-callback-error*)
+                (ssl-signal-error handle func error nbytes))))))))
 
 (declaim (inline nonblocking-ssl-funcall))
 (defun nonblocking-ssl-funcall (stream handle func &rest args)
```

avodonosov:
@zellerin , I would probably even signal an error instead of a warn. But how to distinguish a string that does not allow binary IO?
Using `stream-element-type`? But Common Lisp allows stream with element-type `character` to support binary IO.

For example, we can read characters and bytes from the same flexi-stream:
```common-lisp
(flex:with-input-from-sequence (stream (make-array 10
                                                   :element-type '(unsigned-byte 8)
                                                   :initial-element (char-code #\a)))
  (let ((flexi-stream (flex:make-flexi-stream stream
                                              :external-format :ascii))
        (byte-array (make-array 2 :element-type '(unsigned-byte 8)))
        (char-array (make-array 2 :element-type 'character)))

    (read-sequence byte-array flexi-stream)
    (read-sequence char-array flexi-stream)

    (list (stream-element-type flexi-stream)
          (read-byte flexi-stream)
          (read-char flexi-stream)
          byte-array
          char-array)))

=> (FLEXI-STREAMS::CHAR* 97 #\a #(97 97) "aa")
```
Speaking of the error handling, assigning a dynamic variable should be ok for CFFI. We already access `*socket*`, why not access another variable. But why the current error text was not enough? The error text we get from OpenSSL error queue includes the stringified condition. For me on CCL that's "The value #<STRING-INPUT-STREAM #x302003C7526D> is not of the expected type (AND CCL::BINARY-STREAM CCL:OUTPUT-STREAM)."

```common-lisp
(with-input-from-string (stream "aaaaaaaaaaaaaaaa")
  (cl+ssl:make-ssl-client-stream stream :unwrap-stream-p nil))
```

> A failure in the SSL library occurred on handle #<A Foreign Pointer #x7FAEAC044740> (return code: 1).
140388320933632:error:01000044:unknown library:unknown library:internal error:cl+ssl/src/bio.lisp:0:Unexpected SERIOUS-CONDITION in the Lisp BIO: The value #<STRING-INPUT-STREAM  #x302003C7526D> is not of the expected type (AND CCL::BINARY-STREAM CCL:OUTPUT-STREAM).

zellerin:
You are probably right, I do not know how to detect a binary stream in a compatible way. And as for the error handling - maybe it is not really necessary once the code is well tested, so maybe let us keep it simply as it is now.

For record, my reasoning was that I can inspect the error parameters (stream in this case) in the debugger - in sbcl I have in ssl
```
A failure in the SSL library occurred on handle #.(SB-SYS:INT-SAP #X7FCF2C85FCE0) (return code: 1).
140527900423744:error:01000044:unknown library:unknown library:internal error:cl+ssl/src/bio.lisp:0:Unexpected SERIOUS-CONDITION in the Lisp BIO: #<SB-SYS:FD-STREAM for "socket 192.168.1.85:42832, peer: 93.184.216.34:443" {1003A28ED3}> is not a binary output stream.
```
which is less informative till I inspect the stream itself.

The other idea was to catch expected errors (timeout...) and let only unexpected errors be thrown to detect problems in the code - use case is not a server that is happy to just log its failures, but an interactive client that would handle known specific error and jump to debugger on unexpected ones. But then again, there are probably no expected failures from the lisp callback code other then timeouts.

PR #134, opened 2021-08-08
Potential fix for LW and Big Sur?

tdrhq:
I'm usually a Linux person, but I've been doing some Lispworks on an M1 mac with Big Sur. When using dexador, a simple command like `(dex:get "https://google.com")` would give me the following error:

```
Error: Foreign function CL+SSL::|%cffi-foreign-function/OPENSSL-ADD-ALL-DIGESTS| trying to call to unresolved external function "OpenSSL_add_all_digests".
```
I'm no SSL expect, far from it. But it seems like the `add_all_digests` function is a no-op, so I'm assuming perhaps the version of libressl that ships with my Mac doesn't have it, and if so it should be safe to not call this function?

cc @jathd who added this call very recently in https://github.com/cl-plus-ssl/cl-plus-ssl/commit/a1583d36e60caa6ba29d6fd09210a6edc5990db4


tdrhq:
@avodonosov  Thank you!

PR #137, opened 2021-08-30
Darwin: Load versioned libssl system dylibs

daewok:
This is a follow on to #115. On my Mac M1, I found that the system libssl dylibs need to be versioned as well.


avodonosov:
Thank you

PR #139, opened 2021-11-18
Add support for GeneralizedTime format in X509 certifcate dates.

jathd:
A certificate's notBefore and notAfter dates can be encoded in one of two
formats: UTCTime or GeneralizedTime. Only UTCTime is currently supported. This
commit adds support for GeneralizedTime.


avodonosov:
Thank you, merged. Another pull request with unit tests would be welcomed.

PR #140, opened 2021-11-20
Add test for date encodings in X509 certificates.

jathd:
Check that CERTIFICATE-NOT-BEFORE-TIME and CERTIFICATE-NOT-AFTER-TIME handle
dates both before and after 2050 (different ASN1 encodings are used for these
two cases).


avodonosov:
Merged, thank you

PR #147, opened 2022-03-06
Add read-timeout support for sbcl

ajberkley:
This isn't really meant to be accepted since it only works on sbcl so far.  It works for both BIO and non BIO way of talking to the ssl library.  I imagine I can do the same thing for CCL and some others.  I'm not sure if we can do any better than assuming that on BIO errors that if the timeout expired that the error was caused by the timeout.

### DIFF
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -35,6 +35,9 @@
            #:ssl-error-initialize
            #:ssl-ctx-free
 
+           #:ssl-timeout
+           #:ssl-stream-read-timeout
+
            #:with-pem-password
 
            #:+ssl-verify-none+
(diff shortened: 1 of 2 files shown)


ajberkley:
Hm.  I didn't read through #69 before doing this.  I think I will want to read through that and make sure we can support that the same way (easy enough I think --- if there is a timeout on the underlying stream, just propagate that up to the timeout at ssl-stream build time.  Or we can check it each call, but that seems silly?

ajberkley:
OK, ignore this.  While this works, the adding a deadline for every byte read is ... slow.

ajberkley:
I will take the same approach as #69 but making sure the current methods people use sb-sys:with-deadline, etc continue to work exactly as they do now.  The changes will introduce a new mechanism to specify timeouts which will signal via an cl-plus-ssl condition, unconnected to the old partially implemented methods.

PR #148, opened 2022-03-07
Support user specified READ-TIMEOUT and WRITE-TIMEOUT and optionally honor underlying stream timeout

ajberkley:
This follows the same scheme as #69, but I addressed your desire for complete backwards compatibility.  No one will see timeouts who did not see timeouts before.  Users can request that underlying stream timeouts be obeyed, or can specify read-timeout and write-timeout explicitly at make-ssl-stream time.

clozure-common-lisp maintains its weird status of throwing an implementation specific error and defaulting to honor-underlying-stream-timeout, because I *think* that that is what is was doing before.  I have not tested the allegro, lispworks, or ccl code at all, but will do so once we agree that I've implemented something reasonable here.

Users of sbcl who are using sb-sys:with-deadline will see no changes to behavior.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -877,13 +877,9 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
        (let ((error (ssl-get-error handle nbytes)))
          (case error
            (#.+ssl-error-want-read+
-            (input-wait stream
-                        (ssl-get-fd handle)
-                        (ssl-stream-deadline stream)))
+            (io-wait stream (ssl-get-fd handle) (ssl-stream-deadline stream) :input))
            (#.+ssl-error-want-write+
-            (output-wait stream
-                         (ssl-get-fd handle)
-                         (ssl-stream-deadline stream)))
+            (io-wait stream (ssl-get-fd handle) (ssl-stream-deadline stream) :output))
            (t
             (ssl-signal-error handle func error nbytes)))))))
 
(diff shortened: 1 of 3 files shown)


ajberkley:
With regards to the tests... on sbcl the read-deadline-nil test seems to have never been able to pass... it works fine now if you specify a read-timeout or honor-underlying-thread-timeout, etc.

I don't know if the write timeouts ever worked on sbcl either --- I didn't change the behavior and they hang when I run them.  I'm not sure why we aren't getting called back.

ajberkley:
Hoping you have a chance to look at this.  I'm happy to answer questions!

avodonosov:
@ajberkley, sorry for the silence.

I will try to review in the coming days.

src/ffi.lisp
```diff
@@ -913,122 +912,99 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
...
-           (when deadline
-             (max 0 (seconds-until-deadline deadline)))))
-    (system:wait-for-input-streams (list (ssl-stream-socket stream))
-                                   :timeout timeout
-                                   :wait-reason "cl+ssl waiting for input")))
+#+clozure-common-lisp
+(defun io-wait (stream fd deadline direction)
+  "Supports old behavior that throws `ccl::communication-deadline-expired. if the underlying stream
+   is the source of the deadline.  New behavior invoked by ssl-stream-read-timeout or write-timeout
+   throws `ssl-timeout'"
```
avodonosov, line 988: I agree that the old behaviour should only be invoked by old interface, and new interface should produce the new behavior. But the `milliseconds-until-deadline` definition above signals the old, CCL-specific condition in all situations, right?

src/streams.lisp
```diff
@@ -500,14 +559,22 @@ may be associated with the passphrase PASSWORD."
...
+
+(defmethod stream-deadline ((stream usocket:stream-usocket))
+  (+ (get-internal-real-time)
+     (* internal-time-units-per-second (usocket:socket-option stream :receive-timeout))))
+
+#+sbcl
+(defmethod stream-deadline ((stream sb-sys:fd-stream))
+  (+ (get-internal-real-time)
+     (* internal-time-units-per-second (sb-impl::fd-stream-timeout stream))))
+
```
avodonosov, line 574: I would not want to support portability code for stream timeout extraction in cl+ssl. Not only because that's a pile of conventionalized impl-specific code, but also because it raises some inconvenient questions. Like why do use use `:receive-timeout` for usocket streams, and not `:send-timeout`? What if the lisp implementation does not support stream timeouts, or supports only read timeouts?

It's better to leave the automatic extraction of timeout from stream for CCL backward compatibility support only. 

The users will have to pass the timeouts explicitly as parameters - if they know how to set the timeout in a platform specific way, it should be easy for them to pass the same values to cl+ssl. 

BTW, maybe a separate small portability library could be created for lisp stream timeouts (probably using https://github.com/edicl/hunchentoot/blob/master/set-timeouts.lisp as the starting point). But that's outside of the cl+ssl scope.

src/streams.lisp
```diff
@@ -470,19 +516,32 @@ ALPN-PROTOCOLS, if specified, should be a list of alpn protocol names such as
...
  your server. KEY is the path to the PEM-encoded key for the server, which
-may be associated with the passphrase PASSWORD."
+may be associated with the passphrase PASSWORD.
+
+READ-TIMEOUT and WRITE-TIMEOUT specify a timeout in seconds before
+which to signal an SSL-TIMEOUT error.  If these are not specified and
+HONOR-UNDERLYING-STREAM-TIMEOUT is true, then the underlying SOCKET
+timeout, if set, will be honored and an SSL-TIMEOUT error will be
+thrown except on CCL where an implementation specific error is thrown
+for backwards compatibility"
```
avodonosov, line 533: We will have to mention (in the docstrings of the two functions and in the documentation of the timeout slots) that the timeout parameters only work for file descriptor BIOs, and LispBIOs behave differently.

src/streams.lisp
```diff
@@ -183,6 +215,7 @@
     (when (plusp fill-ptr)
       (unless handle
         (error "output operation on closed SSL stream"))
+      (set-deadline-from-timeout stream :output)
```
avodonosov, line 218: This approach, where every Gray Streams generic functions computes a deadline and stores it as the stream slot, which is later used to decide the timeout expiration, worries me somewhat. It introduces a requirement, that every Gray Stream generic function that my call `ensure-ssl-funcal` has to remember computing and setting the fresh deadline. I worry that in the future code modifications some code path may miss setting new deadline, and the old deadline will be in effect, signalling unnecessary error.

Your approach is probably motivated by the desire to apply the timeout to the full gray stream function, as opposed to smaller scope "any piece of data received". However, even in the current version of the code, the `stream-write-sequence` implementation can apply timeout to chunks of data, instead of the whole data.

I see the following ways to make the implementation more stable in the face of the future changes:
- instead of a stream slot, make the deadline a local variable in the `io-wait` or `ensure-ssl-funcall`, thus shifting the timeout semantics from "the whole stream function completion" to "any data received"
- introduce a macro `(with-deadline-from-timeout (stream) ...)` that will restore the `deadline` field to special value `none` upon exit, and upon enter will only modify the `deadline` slot if it's current value is `none`, which will make nested calls to `(with-deadline-from-timeout (stream)...) a no-operation. It will allow to apply the timeout to the whole `write-sequence`, while still having a timeout for `force-output` in case it's called by the application directly.

The choice depends on which timeout semantics is preferred. 

What do you think, what is better: timeout means the time of full completion of a function, like `read-sequence`, or to mean the time to wait for any data?

FWIW, the SBCL's function `socket-make-stream ` uses the latter behavior.

src/streams.lisp
ajberkley: Agreed, this was an unnecessary feature addition.

src/streams.lisp
ajberkley: I agree things are nicer with "timeout is the time to wait for any data".  I think your worry and suggestions are both good and will do them.

PR #149, opened 2022-03-18
ALPN (Application-Layer Protocol Negotiation) support (client-side)

zellerin:
This PR adds client-side support for ALPN:
- Client can specify what protocols it supports with a parameter to make-ssl-client-stream,
- Client can get what protocol, if any, was accepted by the server with new function.

Any comments/opinions on that, apart from "tests needed"? (I can try, suggestions how welcome) and "Server side support would be nice" (I am not ready for that)?

Example:
```
(usocket:with-client-socket (socket stream "www.example.com" 443)
	(cl+ssl:get-selected-alpn-protocol
              (cl+ssl:make-ssl-client-stream stream :alpn-protocols '( "h2" "http/1.1"))))
```
-> `"h2"`

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -399,6 +399,11 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
 (define-ssl-function ("SSL_CTX_free" ssl-ctx-free)
     :void
   (ctx ssl-ctx))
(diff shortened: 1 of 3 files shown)


avodonosov:
Thank you. Would be good to add that to the doc - index.html.

PR #150, opened 2022-03-19
Documentation (index.html) and tests for client-side alpn

zellerin:
Documentation and tests for #149.

The tests were tested locally only, not across full spectrum of Lisps and SSL versions.

### DIFF
--- a/cl+ssl.test.asd
+++ b/cl+ssl.test.asd
@@ -28,4 +28,4 @@
                  (:file "verify-hostname")
                  (:file "badssl-com")
                  (:file "bio")
-		 (:file "alpn")))))
+                 (:file "alpn")))))
--- a/cl+ssl.test.asd
+++ b/cl+ssl.test.asd
@@ -27,4 +27,5 @@
                  (:file "fingerprint")
                  (:file "verify-hostname")
                  (:file "badssl-com")
-                 (:file "bio")))))
+                 (:file "bio")
+		 (:file "alpn")))))
--- a/index.html
+++ b/index.html
@@ -313,6 +313,10 @@
       Change this variable if you want the previous behaviour.
     </p>
     <p>
+      <div class="def">Function CL+SSL:GET-SELECTED-ALPN-PROTOCOL (ssl)</div>
+      Return ALPN protocol selected by server, or NIL if none was selected. SSL is the client ssl stream returned by <tt>make-ssl-client-stream</tt>.
+    </p>
+    <p>
       <div class="def">Function CL+SSL:USE-CERTIFICATE-CHAIN-FILE (certificate-chain-file)</div>
       Loads a PEM encoded certificate chain file <tt>certificate-chain-file</tt>
       and adds the chain to global context. The certificates must be sorted 
--- a/index.html
+++ b/index.html
@@ -237,7 +237,7 @@
       If <tt>auto-free-p</tt> is true the context is freed using <tt>ssl-ctx-free</tt> before exit.
     </p>
     <p>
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -292,6 +292,11 @@ <h3>API</h3>
       to choose certificate for right domain. Also the <tt>hostname></tt> is used for
       hostname verification if verification is enabled by <tt>verify</tt>.
     </p>
+    <p>
+      <tt>alpn-protocols</tt> if specified, should be a list of alpn protocol
+      names that would be offered to the server. Protocol selected by the server
+      can be retrieved by <tt>get-selected-alpn-protocol</tt>.
+    </p>
```
avodonosov, line 299: It would be good also to add description of the `cl+ssl:get-selected-alpn-protocol`

cl+ssl.test.asd
```diff
@@ -27,4 +27,5 @@
                  (:file "fingerprint")
                  (:file "verify-hostname")
                  (:file "badssl-com")
-                 (:file "bio")))))
+                 (:file "bio")
+		 (:file "alpn")))))
```
avodonosov, line 31: Could you please indent as other files above

test/alpn.lisp
```diff
@@ -0,0 +1,18 @@
...
+(in-suite :cl+ssl.alpn)
+
+(test alpn-client
+  (when (cl+ssl::openssl-is-at-least 1 0 2)
+    (flet ((test-alpn-result (target proposed expected)
+	     (usocket:with-client-socket (socket stream target 443)
+	       (is
+		(equal expected
+		       (cl+ssl:get-selected-alpn-protocol
+			(cl+ssl:make-ssl-client-stream stream :alpn-protocols proposed)))))))
```
avodonosov, line 15: strange indentation (at least in the github web UI)

avodonosov (changes requested):

test/alpn.lisp
zellerin: Tabs problems, in both cases. Will fix.

index.html
zellerin: Ok.

avodonosov:
Merged. thank you.

avodonosov:
@zellerin, somehow the alpn test fails on ABCL, http://cl-test-grid.appspot.com/blob?key=6xx13j9zqd

zellerin:
@avodonosov 
looks like an issue in test itself, the socket must be binary for abcl to work:
```
             (usocket:with-client-socket (socket stream target 443
                                                 :element-type '(unsigned-byte 8))
```
The test passes after the change when run locally. Will prepare patch later this weekend if needed/confirmed.

avodonosov:
@zellerin , one more question.

When I run the testsuite loally, I have the alpn test failed like this:
```
 Did 67 checks.
    Pass: 66 (98%)
    Skip: 0 ( 0%)
    Fail: 1 ( 1%)

 Failure Details:
 --------------------------------
 ALPN-CLIENT in CL+SSL.ALPN []: 
      Unexpected Error: #<CL+SSL:SSL-ERROR-VERIFY #x302003DA4ECD>
SSL verify error: 20 X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY.
 --------------------------------
```

Do you have this now?

avodonosov:
My failure was on CCL

avodonosov:
Turned out my failure was only happening when I use OpenSSL binaries I build myself for CI and does not happen when I use the system installed OpenSSL.

Solved by configuring OpenSSL to use the trusted certs from the directory where my system stores them and which is used by the system OpenSSL:
```common-lisp
(cl+ssl:ensure-initialized) ; needed to set the *ssl-global-context* 
(cl+ssl::ssl-ctx-set-verify-location cl+ssl::*ssl-global-context* "/etc/ssl/certs/")
```

PR #157, opened 2022-04-17
Add READ-TIMEOUT and WRITE-TIMEOUT support

ajberkley:
Throws an SSL-TIMEOUT error on timeout.

The behavior applies to both full operations (read-sequence) or individual operations making up the calls.

Continue to support MCL/CCL underlying stream deadlines which when triggered throws an internal CCL error.

Continue to support slot ssl-stream-deadline in case people were using it (despite it not being exported).

I believe this addresses your worries about deadlines and timeouts --- I use a 'nestable' with-timeout macro that is used in ensure-ssl-funcall as well as in stream-read-sequence and stream-write-sequence.  This means the timeout applies to the whole of the read-sequence operation.  In future changes, forgetting to add the with-timeout wrapper means that the timeout applies to underlying individual io calls.  This is implemented as you suggested by having ssl-stream-operation-started always be :not-in-progress until a with-timeout macro is hit, and then it is set to something.  Once it is set, further recursive calls do not update it. There should be no way to get a spurious deadline where there shouldn't be one as it is cleared in an unwind-protect when leaving the outer-most with-timeout call.

The only thing I find aesthetically wrong about the solution is that the macro is used both in ffi.lisp and streams.lisp and lives in ffi.lisp but knows stuff about streams.  If you have any suggestions, I can do make them.

I tested this on SBCL and dexador and it works fine.  I did not test on any other implementation... I don't have any setup, and as you can guess I have very limited time for these endeavors.  I would appreciate any help.

### DIFF
--- a/src/ffi.lisp
+++ b/src/ffi.lisp
@@ -880,26 +880,55 @@ Note: the _really_ old formats (<= 0.9.4) are not supported."
 ;;;
 (defvar *socket*)
 
+(defmacro with-timeout ((stream read-deadline write-deadline) &body body)
+  "Handle read-timeout and write-timeout requests within the body.  If
+   no other operation is in progress, updates
+   ssl-stream-operation-started to be the current time and then
+   executes body.  Binds READ-DEADLINE and WRITE-DEADLINE to a
+   deadline in internal time units which should be obeyed (if either
+   is nil then that timeout type is ignored).  On error, resets
+   operation started if it had been set by this call."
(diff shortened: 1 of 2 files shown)


ajberkley:
I know life is busy and all, but is there any chance this is going to get looked at?

ajberkley:
@avodonosov

avodonosov:
@ajberkley , sorry for the pause. Yes, I want to finish this one day.

Issue #158, opened 2022-05-23
Getting client certificates seems to fail (null pointer?)


cage2:
Hi!

I am trying to get the fingerprint of a client certificate from an ssl-stream but with no luck so far. 
I am using the function `ssl-stream-x509-certificate` but i get an error that says that the client sent no certificate at all, inspecting the stacktrace seems to me that the function above returned a null pointer, in fact. I am reasonably sure that the test client I am using (the `openssl` command line program with `-cert` and `-key` with file in pem format) is sending the certificate. I was even able to actually send some data from the client to my server after i bypassed the certificate fingerprint function. I wonder if I am using the correct function. In another program i was able, with the code you can find in the link below:

https://notabug.org/cage/tinmop/src/master/src/x509.lisp#L60

to get the fingerprint of the server's certificate, i assumed the same function worked for fetching the client's certificate (when provided of course! :)).

Sorry i have not the actual code to show, i can provide it in the next hours if needed.

Maybe is useful to know that the client certificate is self signed.

Thanks in advance!
C.

avodonosov:
I neer did that.

All I can say is that `ssl-stream-x509-certificate` simply calls OpenSSL's ` SSL_get_peer_certificate`  or since version 3.0.0 `SSL_get1_peer_certificate`.

Checking the OpenSSL docs may help.

avodonosov:
> A client will only send a certificate when explicitly requested to do so by the server (see [SSL_CTX_set_verify(3)](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify.html)).

avodonosov:
So, it seems you need your server streams or context to use the `SSL_VERIFY_PEER`

cage2:
Hi @avodonosov 

> So, it seems you need your server streams or context to use the `SSL_VERIFY_PEER`

Thank Anton! Again you have given me the right suggestion! :)

Exactly the constant above is needed when creating the context to get the client certificate.

The only caveat is that, if the certificate is self signed, you have to specify a callback for the `:verify` argument of `MAKE-CONTEXT` that always returns `1` otherwise an error is signaled.

I would like to add an example for this procedures to this project  as i think that someone could benefit from it if a client verification is needed (i am thinking about a gemini server, for example).

Let me know if my proposal is worth, in the meanwhile i am closing this and, obviously, thank you very much!
C.

avodonosov:
[...]

PR #159, opened 2022-05-25
Added and example to shows a possible way to use client certificates

cage2:
Hi!!

I tried to add an example as we discussed in #158. Hope this code is clear enough!

Feel free to comment or propose improvements if you want, of course!

Bye!
C.

### DIFF
--- a/client-certificates-example.lisp
+++ b/client-certificates-example.lisp
@@ -1,0 +1,94 @@
+;;;; The code contained in this file implements a trivial server and a
+;;;; client  that connects  to the  former and  provide a  self signed
+;;;; certificate.  The server  is  able to  implement  a procedure  to
+;;;; reject or  accept the  client connection,  based on  the client's
+;;;; certificate, and using some  form of authentication.  For example
+;;;; matching  the   certificate  fingerprint   with  a   database  of
+;;;; certificates stored on disk, for example.
+
+;; The key points here are:
+
+;; For the server
+;; - create a context using :verify-mode cl+ssl:+ssl-verify-peer+
+
+;; Optional only if you plan to use self signed client certificates
+
+;; - write  a CFFI callback that  always return 1 (one)  to bypass the
+;; certificate authentication against  a certification authority; pass
+;; the  symbol  of  this  callback   to  key  parameter  :verify  when
+;; generating the context
+
+;; For the client
+
+;; - pass  certificate and key when  generating the stream as  you did
+;; for the server
+
+(asdf:make "cl+ssl")
+
+(asdf:make "bordeaux-threads")
+
+(asdf:make "trivial-sockets")
+
+(cffi:defcallback no-verify :int ((ok :int) (ctx :pointer))
+  (declare (ignore ok ctx))
+  1)
+
+(defun hash-array->string (array)
+  (let ((res ""))
+    (loop for i across array do
+      (setf res
+            (concatenate 'string
+                         res
+                         (format nil "~2,'0x" i))))
+    res))
+
+(defun start-trivial-server (port certificate key)
+  "Start a  trivial server listening  on `PORT' using  the certificate
+and  key   stored  in  the   file  pointed  by  the   filesystem  path
+`CERTIFICATE' and `KEY' respectively"
+  (format t "~&SSL server listening on port ~d~%" port)
+  (bt:make-thread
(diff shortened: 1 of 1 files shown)


avodonosov:
@cage2 , thank you, merged the pull request. Also moved the file to a new examples/ directory and did micro changes.

If you have example commands to generate the certificates, we could add them to the comments. Thus lowering the barrier for the user to run the examples.

Also, I think of alternatives to the approach demonstrated here and would like to know your opinion.

In your example, in case the server decides to reject the client cert, I assume you mean to return HTTP level error, like 403.

As an alternative to this, the cert validation logic could be placed inside the validate callback . In this case the connection would be terminated by the server during TLS handshake.

So what is the best practice, from security, usability and established patterns in the industry? HTTP level error, or TLS level error?

As an extension to this alternative, one could prepare a file system directory containing all the client certs the server should trust, or containing some root certificate, that is used to issue all client certificates. And then configure the CTX to trust those certs using SSL_CTX_load_verify_location. That way the user would completely rely on OpenSSL for verification, instead of writing his own verification code. (But in case of root cert approach, extra care will be needed to support quick revocation of client certs - maybe maintain a Certificate Revocation List endpoint and set the CRL verification flags for OpenSSL. When listing individual client certs, either on file system, or in DB, we only need to make sure that after revocation the cert is removed from that storage and the server notices the change).

What is your opinion on the above questions?

cage2:
On Tue, May 31, 2022 at 07:43:10PM -0700, Anton Vodonosov wrote:

Hi!

> @cage2 , thank you, merged the pull request.

Thanks to you for reviewing the patch!

> Also moved the file to a new examples/ directory and did micro changes.

Improvements, i would say! :)

> If you have example commands  to generate the certificates, we could
> add them to the comments. Thus  lowering the barrier for the user to
> run the examples.

Good idea!  I'll prepare a  patch with the  openssl command, it  is so
simple that the command line could be added ina comment on top of the
file.

> Also, I think of alternatives  to the approach demonstrated here and
  would like to know your opinion.
>  In your  example, in case the  server decides to reject  the client
> cert, I assume you mean to return HTTP level error, like 403.
>  As  an alternative  to this,  the  cert validation  logic could  be
> placed inside  the validate callback  . In this case  the connection
> would be terminated by the server during TLS handshake.

> So  what  is  the  best   practice,  from  security,  usability  and
> established patterns in the industry? HTTP level error, or TLS level
> error?

Well, I am very far from an expert! :)

I am using a  code similar to the one i proposed in  the example for a
simple  server  that  receives,  from  trusted  clients,  a  bunch  of
serialized sexp. To  verify the identity of the client  i just get the
fingerprint of the client's certificate:

(cl+ssl:ssl-stream-x509-certificate ssl-stream)

And matching the fingerprint with the ones contained in a database.

This was my idea to authorize clients,  honestly i do not know if there
was a better way but i suspect such better way exists! :)

For  example  i  wonder  if  such   matching  could  be  made  in  the
verify-callback as,  in my opinion,  the earlier the rejection  of the
client is done the better (even from a bandwidth perspective).

> As an extension to this alternative, one could prepare a file system
> directory containing all  the client certs the  server should trust,
> or  containing some  root certificate,  that  is used  to issue  all
> client certificates. And then configure the CTX to trust those certs
> using   SSL_CTX_load_verify_location.  That   way  the   user  would
> completely rely on OpenSSL for  verification, instead of writing his
> own verification code.

I think this is the better way I was writing above!

> (But in  case of root  cert approach, extra  care will be  needed to
>support  quick  revocation  of  client   certs  -  maybe  maintain  a
>Certificate  Revocation List  endpoint and  set the  CRL verification
>flags for  OpenSSL. When listing  individual client certs,  either on
>file  system,  or  in DB,  we  only  need  to  make sure  that  after
>revocation  the cert  is removed  from  that storage  and the  server
>notices the change).  What is your opinion on the above questions?

To me the  custom CA approach could be complicated  for some scenarios
like a gemini server (for example) that do not uses CA at all.

I am going  to use your suggestion  and will try to  configure the SSL
library to  check the  certificates from an  user defined  file system
directory. This changes could even makes the lives of the user simpler
as  they have  just to  send the  certificate to  sysadmin instead  of
extract the fingerprint! I am thinking  i would upgrade the example to
use this mechanism.

Thanks again for the suggestion!
C.

PR #162, opened 2022-06-18
Fix failing alpn test on abcl.

zellerin:
See #150. Tested on ABCL and sbcl, but should be universal.


avodonosov:
Thank you, merged

PR #178, opened 2023-04-15
added libcrypto-3 for Windows

gendl:


avodonosov:
Ok, I will add this, and the same for libssl.

Does it work for you in the presented way, with new libcrypto but old libssl?

avodonosov:
Added this libraries. Thank you.

gendl:
Hi  @avodonosov,
 
That `libcrypto-3-x64.dll` for the libssl definition in `reload.lisp` was a typo or mistake on my part. Indeed it did appear to work for me (probably by some kind of accident), but clearly that libssl definition should have `libssl-3...` not `libcrypto-3...` 
   
 I will do some more builds & testing later today with the latest master and will report if my tests are not still working as expected. 
 
 Thanks for catching that typo and making the fix, and sorry for the confusion. 

 Dave

PR #180, opened 2023-07-28
Fix issue of having both x86-64 and arm64 Homebrew installation and running lisp with arm64 architecture

svetlyak40wt:
When you are having both x86-64 and ARM Homebrew, and running LISP with arm architecture, attempt to load cl+ssl fails with error:

```
Unable to load foreign library (LIBCRYPTO).
  Error opening shared object "/usr/local/Cellar/openssl@1.1/1.1.1s/lib/libcrypto.1.1.dylib":
  dlopen(/usr/local/Cellar/openssl@1.1/1.1.1s/lib/libcrypto.1.1.dylib, 10): no suitable image found.  Did find:
        /usr/local/Cellar/openssl@1.1/1.1.1s/lib/libcrypto.1.1.dylib: mach-o, but wrong architecture
        /usr/local/Cellar/openssl@1.1/1.1.1s/lib/libcrypto.1.1.dylib: stat() failed with errno=4.
   [Condition of type CFFI:LOAD-FOREIGN-LIBRARY-ERROR]
```

This is because [this code](https://github.com/cl-plus-ssl/cl-plus-ssl/blob/13d824e27cf7f6085086458daea1514b605b3980/src/reload.lisp#L105-L106) tries to load x86 version of the library first and fails.

My patch fixes this by removing one or another path depending on current Lisp architecture.

### DIFF
--- a/src/reload.lisp
+++ b/src/reload.lisp
@@ -102,8 +102,8 @@ sudo rm /usr/local/lib/libcrypto.dylib /usr/local/lib/libssl.dylib
 
     ((:and :darwin :cl+ssl-macports-found) "/opt/local/lib/libcrypto.dylib")
     ((:and :darwin :cl+ssl-fink-found) "/sw/lib/libcrypto.dylib")
-    ((:and :darwin :cl+ssl-homebrew-found) "/usr/local/opt/openssl/lib/libcrypto.dylib")
-    ((:and :darwin :cl+ssl-homebrew-arm64-found) "/opt/homebrew/opt/openssl/lib/libcrypto.dylib")
+    ((:and :darwin :x86-64 :cl+ssl-homebrew-found) "/usr/local/opt/openssl/lib/libcrypto.dylib")
+    ((:and :darwin :arm64 :cl+ssl-homebrew-arm64-found) "/opt/homebrew/opt/openssl/lib/libcrypto.dylib")
     ((:and :darwin :cl+ssl-personalized-install-found) "/usr/local/lib/libcrypto.dylib")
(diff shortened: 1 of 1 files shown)


avodonosov:
@svetlyak40wt, does mac have 32-bit version? I mean, shouldn't we do `(or :x86 :x86-64)` instead of simply `:x86-64`? Or maybe even `(not :arm64)`?

svetlyak40wt:
I don't know. Maybe some old OSX versions can run on some old hardware?

I think we could make `(or :x86 :x86-64)` just to be sure.

avodonosov:
The Homebrew installation doc has a footnote 1, saying "For 32-bit or PPC support see [Tigerbrew](https://github.com/mistydemeo/tigerbrew)"

https://docs.brew.sh/Installation#1

And the Tigerbrew install script says "This script installs to /usr/local only":
https://github.com/mistydemeo/tigerbrew/blob/b7ab6258bd595fa0568703326326abebb272e288/install#L2C1-L2C1

But I do not understand whether it installs to /usr/local/opt/openssl/lib/

svetlyak40wt:
I don't have such old hardware to check.

svetlyak40wt:
Probably we could merge as `(or :x86 :x86-64)`, fix at least for modern systems and hope it will not break more rare old one?

avodonosov:
I lean towards `(:not :arm64)` PPC is also mentioned for Tigerbrew.

I hope add this with a coment and merge in the evening today.

svetlyak40wt:
Cool! Thank you!

avodonosov:
Merged, thank you.

PR #192, opened 2025-07-07
Fix memory leak when with-new-ssl exits from non-error condition

kchanqvq:
Some (usually implementation-defined) non-error condition can causes non local exit, e.g. breakpoints and SBCL timeout/deadline. The previous implementation only calls `ssl-free` on error conditons, which will leak memory in such cases. This patch fixes this.

### DIFF
--- a/src/streams.lisp
+++ b/src/streams.lisp
@@ -379,13 +379,15 @@ MAKE-CONTEXT also allows to enab/disable verification."
 (defmacro with-new-ssl ((var) &body body)
   (alexandria:with-gensyms (ssl)
     `(let* ((,ssl (ssl-new *ssl-global-context*))
-            (,var ,ssl))
+            (,var ,ssl)
+            (normal-exit nil))
        (when (cffi:null-pointer-p ,ssl)
          (error 'ssl-error-call :message "Unable to create SSL structure" :queue (read-ssl-error-queue)))
-       (handler-bind ((error (lambda (_)
-                               (declare (ignore _))
-                               (ssl-free ,ssl))))
-         ,@body))))
+       (unwind-protect
+            (multiple-value-prog1 (progn ,@body)
+              (setq normal-exit t))
+         (unless normal-exit
+           (ssl-free ,ssl))))))
 
 (defvar *make-ssl-client-stream-verify-default*
   (if (member :windows *features*) ; by trivial-features


avodonosov:
Hi, this solution seems ok.

Quesgion: have you experienced this leak in practice, or that's just a theoretical worry.

BTW, the SBCL timeouts/deadlines would be covered if the handler-bind was set for serious-condition instead of error. Breakpoints are not non-local exits, because the execution is just paused - the call stack is not unwound.

Your solution is more reliable, probably, than simply switching to serious-condition. Although not 100% reliable - the timeout may happen after the normal-exit is set to t, for exampke during the execution of the cleanup form of the unwind-protect.

kchanqvq:
Thanks for reviewing!

> Question: have you experienced this leak in practice, or that's just a theoretical worry?

Yes, this bug is discovered entirely due to a mysterious leak on one of my server. The process has a polling loop that sets a deadline, which occasionally triggers. The process would run for months slowly leaking memory then dies from OOM.

> the timeout may happen after the normal-exit is set to t, for example during the execution of the cleanup form of the unwind-protect.

Indeed. In practice this is already much better: for example, my server process that sets deadline in polling loop is now completely free from memory leak because there's 0 chance another deadline happens (there's about 1 second) during cleanup form.

In theory to guard from this I'm afraid currently the only possible solution will be implementation specific (using e.g. `sb-ext:without-interrpts`), and I'm not aware of any `trivial-*` portability library for this. Maybe we should write a `trivial-without-interrupts`?).

In any case, I suggest maybe let's merge this for now, and the interrupted cleanup form edge case can be dealt with later/in a separate issue!

avodonosov:
@kchanqvq , I've committed the fix based on your solution (made a little step further to introduce a reusable `abort-protect` macro). Have a look if you're interested.

avodonosov:
>> the timeout may happen after the normal-exit is set to t, for example during the execution of the cleanup form of the unwind-protect.

kchanqvq:
> [...] there's 0 chance another deadline happens (there's about 1 second) during cleanup form

I meant not another deadline, but he first deadline happens to occur during cleanup form.

But I was wrong, this problem does not exist at all - SBCL deadlines only happen during blocking operations (input/output, sleep, etc); and don't happen during execution of regular code, like we have in the cleanup forms.

Question: I guess on your server deadlines happen during `make-ssl-client-stream`, and your covered `make-context` just for for sure, although they don't produce deadlines in your practice, right? I assume this because `make-ssl-client-stream` really preforms network communication with `ssl-connect` and can have a deadline with unresponsive server. While `make-context` is a fully local operation which I think should never face any deadlines. Also I guess you are not calling make-context repeatedly. 

Probably the root cause is that `make-ssl-client-stream` does too much and thus is a sub-optimal API. If it was split into two operations: 1) make a stream,  2) connect it, the user ;library user could preform the the cleanup himself (by closing the stream, which he needs to do anyways - and that contains `ssl-free` inside)

Issue #120, opened 2021-03-06
Memory leak of stream buffers? (on CLISP)


avodonosov:
The `ssl-stream` class [allocates](https://github.com/cl-plus-ssl/cl-plus-ssl/blob/51885381423cfbdbc3fb3b2eef743f08045c6907/src/streams.lisp#L43) input and output buffers, but they are not de-allocated anywhere.

The [`close`](https://github.com/cl-plus-ssl/cl-plus-ssl/blob/51885381423cfbdbc3fb3b2eef743f08045c6907/src/streams.lisp#L71)
 method seems to be the right place to de-allocate the buffers.

In most lisps the buffer is [created](`https://github.com/cl-plus-ssl/cl-plus-ssl/blob/51885381423cfbdbc3fb3b2eef743f08045c6907/src/ffi-buffer.lisp#L8`) with `cffi:make-shareable-byte-vector` what should be garbage collected I think,  but in CLISP the buffer is [created](https://github.com/cl-plus-ssl/cl-plus-ssl/blob/51885381423cfbdbc3fb3b2eef743f08045c6907/src/ffi-buffer-clisp.lisp#L48) with `(cffi-sys:%foreign-alloc size)`

svetlyak40wt:
Interesting. On Lispworks I also experience memory leaks when accessing HTTPS sites via dexador. Dexador uses CL+SSL [like this](https://github.com/fukamachi/dexador/blob/master/src/backend/usocket.lisp#L465-L489).

On Lispworks this version of the `make-buffer` is used:

```
(defun make-buffer (size)
  (cffi:make-shareable-byte-vector size))
```

and `make-shareable-byte-vector` just creates a statically allocated array:

```
(defun make-shareable-byte-vector (size)
  "Create a shareable byte vector."
  #+(or lispworks3 lispworks4 lispworks5.0)
  (sys:in-static-area
    (make-array size :element-type '(unsigned-byte 8)))
  #-(or lispworks3 lispworks4 lispworks5.0)
  (make-array size :element-type '(unsigned-byte 8) :allocation :static))
```

I didn't find in the LW documentation any information about the need for manual deallocation of such arrays.

avodonosov:
Yes, LW docs [say](http://www.lispworks.com/documentation/lw70/LW/html/lw-1272.htm) the arrays created in static area are garbage collected.

[...]

PR #196, opened 2025-11-26
Fix memory corruption and leaks on clisp

vibs29:
----
Files
  ffi-buffer.lisp
  ffi-buffer-clisp.lisp
  streams.lisp
  random.lisp
  x509.lisp

----
Memory corruption bug on clisp

s/b-replace
b/s-replace
had bugs that could cause them to miscalculate the buffer's end as being
beyond its boundary, or miscalculate the number of bytes to copy if the
buffer's end was specified but the sequence was smaller.
All callers of s/b-replace happened to pass arguments that didn't trigger
its bugs.
But one caller of b/s-replace (namely stream-write-sequence) could
legitimately call it in a way that did trigger one of its bugs.
E.g. if the buffer was smaller than the sequence, it would corrupt memory
by writing beyond the buffer's bounds.

I have fixed all the bugs, which were in s/b-replace and b/s-replace.

----
Performance

For clisp:

b/s-replace also copies less.
The old version always called subseq, which copies.
The new version copies only if the source seq is not a vector.

s/b-replace is not expected to allocate memory proportional to its
input arrays.
But due to its call to memory-as, it did.
Now it doesn't. It allocates O(1) memory, regardless of input array sizes.

b/s-replace also allocates O(1) memory now.

For all lisps:

stream-read-sequence
I have made this clearer.

stream-write-sequence
I have rewritten this to be clearer.
And faster. The old version could flush a non-full stream.
As a pathological case,
  writing 1 byte, then 2048 times writing 2049 bytes, would cause 4096
  flushes.
Now that will cause only 2049 flushes.

----
Memory leak on clisp

There was also a memory leak because foreign buffers were allocated but
never freed. I have fixed this by extending the buffer API and having
all callers of make-buffer also release the buffer when finished with it.

### DIFF
--- a/src/ffi-buffer-clisp.lisp
+++ b/src/ffi-buffer-clisp.lisp
@@ -34,6 +34,31 @@
 ;;;; on clisp under this patch. But you weren't relying on cl+ssl
 ;;;; internals anyway, now were you?
 
+#|
+2025-11-22
+Comments:
+1. MEMORY-AS copies. So the 60x speedup from 15 minutes to 15 seconds for
+   large downloads could not have been due to avoiding copying the buffer
+   because this S/B-REPLACE also copies the buffer (just once instead of
+   twice as the original wptvd would do). Such a large speedup was likely
+   due to copying via a single foreign call to MEMORY-AS instead of one
+   foreign call per element via %MEM-REF.
+2. streams.lisp's STREAM-LISTEN and STREAM-READ-BYTE would however indeed
+   have been sped up 100 or 1000x due to copying only one byte instead of
+   the whole buffer.
+3. The solution above claims to break wptvd's abstraction. In fact, wptvd
+   was not abstract enough and the solution increases (and improves) its
+   abstraction. Making the buffer into a proper abstract data type so users
+   of its instances don't pry into its internal implementation improves
+   the wptvd abstraction. ffi-buffer.lisp and ffi-buffer-clisp.lisp should
+   be moved into cffi, replacing what cffi already has.
+Improvements: In this updated version of the file:
+- All memory leaks are fixed.
+- All array boundary miscalculations, which would cause crashes, are fixed.
+- S/B-REPLACE and B/S-REPLACE now require O(1) memory as expected of them.
+- Data copying is reduced.
+|#
+
 (in-package :cl+ssl)
 
 (defclass clisp-ffi-buffer ()
@@ -49,6 +74,12 @@
                  :size size
                  :pointer (cffi-sys:%foreign-alloc size)))
 
+(defun release-buffer (buf)
+  (let ((addr (clisp-ffi-buffer-pointer buf)))
+    (when (ffi:validp addr)
+        (ffi:foreign-free addr)
+        (setf (ffi:validp addr) nil))))
+
 (defun buffer-length (buf)
   (clisp-ffi-buffer-size buf))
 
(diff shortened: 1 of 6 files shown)


avodonosov:
re #120

src/x509.lisp
avodonosov: What motivates the changes in this file, and what is the changes nature? Note, `decode-certificate` and `decode-certificate-from-file` are public functions.

avodonosov:
How to avoid unnecessary changes in unmodified files is a secondary question. We will solve that after we agreed on the final version. 

But first I need to understand all the changes. I haven't digested your branch yet.

The comments in src/ffi-buffer-clisp.lisp say you got a significant speed up and attribute that to "copying via a single foreign call to MEMORY-AS instead of one foreign call per element via %MEM-REF.' Could you point where this mem-ref per element happens in the old code?

What whould be good to have is self-contained test cases demonstrating the bug in the old code and pass with the new code, and also covering all branches of the copying code changed and introduced. Do you see a sufficienty easy and practical way to implement such tests?

Are you open go have a call to help me understand the pull request?

vibs29:
I'm unfamiliar with the bio code. I've taken a quick look at the write-puts test. I don't think what I've done will directly fix that. However, remembering that stream-write-sequence would on clisp illegally overwrite arbitrary parts of the C memory, if any of the tests called stream-write-sequence then all subsequent behaviour of that process is unpredictable. It'll certainly be worth trying the test suite of 163 to see if its problem has disappeared with this patch. And it's not worth reasoning about any misbehaviour on clisp prior to applying this patch. cl+ssl on clisp is dangerous and should never be used without this patch.

vibs29:
This is how I validate the last commit 45c439f "Stabilise performance for byte-sized operations". `*url*` needs editing first. Expect output of about one dot, lots of dashes, then a string form of the http response.
```
; insert the following traces into cl+ssl's streams.lisp to ensure it is called.
; stream-listen
;   (princ #\.) (finish-output)
; stream-read-byte
;   (princ #\-) (finish-output)

(defparameter *url* "https://...")

(defun test-byte ()
  "Exercises stream-listen and stream-read-byte"
  (multiple-value-bind (stream code headers uri stream2 must-close reason)
                       (drakma:http-request *url* :want-stream t)
    (unwind-protect
      (let ((str (chunga:chunked-stream-stream
                   (flexi-streams:flexi-stream-stream stream))))
        (print (list 'str-type (type-of str)))
        (do () ((listen str)))
        (princ
          (with-output-to-string (o)
            (handler-case
                (do ((b (read-byte str nil :eof)
                        (read-byte str nil :eof)))
                    ((eq b :eof))
                  (write-char (code-char b) o)) ;assume character
              (cl+ssl::ssl-error-ssl (e)
                (print e)
                ;SSL_get_error specifies not to call SSL_shutdown
                ;(but now your test leaks C buffer memory.)
                ;really, the error should be trapped by stream-read-byte
                (setf must-close nil))))))
      (when must-close
        (close stream))))
  (values))
```

src/streams.lisp
```diff
@@ -112,7 +116,7 @@
           (ssl-stream-peeked-byte stream)
         (setf (ssl-stream-peeked-byte stream) nil))
       (handler-case
-          (let ((buf (ssl-stream-input-buffer stream))
+          (let ((buf (ssl-stream-input-buffer-small stream))
```
avodonosov: Why reading into a single-byte buffer is better than reading one byte into the big buffer in stream-listen and stream-read-byte?

avodonosov: Sorry I missed the commit message (I looked in the pull request conversation and in the code comments before asking though).

I see now in the commit message you adress that question.

src/ffi-buffer-clisp.lisp
```diff
@@ -58,40 +89,55 @@
...
 
-(declaim
- (inline calc-buf-end))
-
-;; to calculate non NIL value of the buffer end index
-(defun calc-buf-end (buf-start seq seq-start seq-end)
-  (+ buf-start
-     (- (or seq-end (length seq))
-        seq-start)))
+(defparameter *mem-max* 1024 "so *-REPLACE require the expected O(1) memory")
```
avodonosov: The `*mem-max*` is intended to limit the allocations done by `memory-as`, right?  Question, if we copy 2048 bytes, calling memory-as twice for 1024 bytes, is it really better than a single call for 2048 bytes? The total amount of allocated memory is the same. If it is garbage collected, collecting two objects may be more work for GC than one object. Is it really beneficial?

Also, in context of cl+ssl, the maximum size of arrays copied with `s/b-replace`, `b/s-replace` is limited by the buffer size, which defaults to 2048 bytes. Not a big difference from the 1024 `*mem-max*` here.

Are you thinking of the `*-replace` functions as more general purpose utilities, that must be prepared to any size of collections copied?

avodonosov:
My comments about the Eventual Design Ideal:

The `cffi:with-pointer-to-vector-data` is not a bad API, it's just intended for a different use-case. Namely, for applications which want to call C functions like `read` or `SSL_read` directly, and have these functions to directly populate a lisp sequence, with zero intermediate copies. On lisp implementations that provide zero copying `cffi:with-pointer-to-vector-data` it is an ideal API for that case: give C function a native foreign pointer to the memory underlying a lisp sequence and then work with the sequence in Lisp using standard sequence functions.

In cl+ssl we have a different arrangement: we read into a buffer and then copy from the buffer to another sequence. So we may want better support in CFFI for this use case without deprecating `cffi:with-pointer-to-vector-data`.

Next, I think there is not need in CFFI to introduce an opaque buffer abstraction.

Applications can just use `foreign-alloc` and `foreign-free` for buffer allocation / release, `men-ref` and `mem-aref` for element-wise access.

The only thing missing is bulk copy between a foreign memory array and lisp sequence  (`b/s-relpace`, `s/b-replace` in terms of cl+ssl). If CFFI introduced functions for that, it will be feature complete for the use case we have in cl+ssl.

A big design question is how generic these bulk copy functions should be with respect to lisp sequence type: support only simple vectors, any vectors, or both vectors and lists. This depends on the possibilities we can reasonably expect from lisp implementations to efficiently provide.

avodonosov:
I undersdod what may be the motivation for opaque buffer abstraction in cffi: if efficient bulk co0ying operations are not available in all lisp, then some lisps may implement buffer using shareable-vector/wptvd and other lisps with foreign-alloc / bulk copy.

PR #197, opened 2025-11-28
Make loadable on clisp without threads

vibs29:
Bordeaux-threads used to be a portability layer over threaded and non-threaded Lisps such that libraries (such as cl+ssl) could write (make-lock) and (acquire-lock) without read conditionals to check whether the Lisp implementation supported threads at all. Bordeaux-threads provided correct no-op definitions for such functions on non-threaded Lisps. At some point after bordeaux-threads-0.8.8 it decided to not load into non-threaded Lisps any more by design.

That meant that users of e.g. clisp without threads could no longer even use Drakma to fetch a web page, (recalling that the web has moved from http to https).

All libraries built on bordeaux-threads, assuming they want to continue supporting non-threaded Lisps, must therefore introduce conditionals around various bordeaux-threads function calls.

This commit makes that change to cl+ssl so that cl+ssl loads and works again on non-threaded clisp even with the newest bordeaux-threads.

Aside: clisp can be built with thread support. However that is not just beta, it is known to have race conditions because some data structures have not yet been made thread-safe. So the only correct variant of clisp is one built without thread support, which is the default build everywhere I've seen.

### DIFF
--- a/cl+ssl.asd
+++ b/cl+ssl.asd
@@ -15,8 +15,8 @@
                :cffi
                :trivial-gray-streams
                :flexi-streams
-               :bordeaux-threads
-               :trivial-garbage
+               (:feature (:or (:not :clisp) :mt) :bordeaux-threads)
+               (:feature (:or (:not :clisp) :mt) :trivial-garbage)
                :uiop
                :usocket
                :alexandria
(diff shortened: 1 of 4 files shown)


avodonosov:
Somebody could create a portability library with API equal to bordeaux-threads and implementing the no-ops for non-threaded lisps, and with fall back to real bordeaux-threads on threaded lisps :)

